PR #24, opened 2019-11-19
Use Quarkus Tools like for vscode and eclipse

maxandersen:


maxandersen:
we now require signoff on PR's ? that’s a shame - so can't use github ui :/ 

yeah, I think Quarkus Tools is a nice way of ensuring consistency of naming and give it a name that is possible to refer to without mixing Quarkus itself up with the tools for Quarkus.

jeffmaury:
Yes that's a @richardfontana requirement.
But he VSCcode extension is named Quarkus even if the title is `Quarkus Tools for Visual Studio Code`

image

maxandersen:
ah sure - i'm just talking about this main page and how its presented on intellij marketplace where I assume this readme eventually will show up too ?

jeffmaury:
No JetBrains marketplace is not as good as VSCode one: the info showed there is embedded in the plugin.xml thus we need duplication there.

jeffmaury:
Replaced by #37

PR #62, opened 2020-03-22
No lsp plugin

jeffmaury:


maxandersen:
what is the hold up on getting this in ? if it avoids the lsp compatibility issues is this not absolute necessary for stability ?

jeffmaury:
I'm in the process of stress testing it then it's going to be merged.

Issue #11, opened 2019-10-17
Add the qds code snippet


jeffmaury:
Add a code snippet for configuring a Quarkus datasource in application.properties

sdaschner:
Added something in #74 for this and for other code snippets

jeffmaury:
Fixed in 1.0.0

PR #74, opened 2020-04-21
Initial live template support

sdaschner:
Added live templates suggested in #11 and others.

Suggested:

- `qds`: JDBC datasource in properties
- `ijt`: Inject a CDI bean
- `conf`: Inject a config property
- `pdm`: Create a CDI producer method
- `asd`: `@ApplicationScoped`
- `rsd`: `@RequestScoped`
- `dep`: `@Dependent`
- `pct`: Create an initializer method
- `pds`: Create an pre-destroy method
- `tn`: `@Transactional`
- `ph`: `@Path`
- `projson`: `@Produces(MediaType.APPLICATION_JSON)`
- `conjson`: `@Consumes(MediaType.APPLICATION_JSON)`
- `getm`: `@GET` method

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/liveTemplates/QuarkusApplicationPropertiesContext.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/liveTemplates/QuarkusApplicationPropertiesContext.java
@@ -1,0 +1,18 @@
+package com.redhat.devtools.intellij.quarkus.liveTemplates;
+
+import com.intellij.codeInsight.template.TemplateContextType;
+import com.intellij.psi.PsiFile;
+import org.jetbrains.annotations.NotNull;
+
+public class QuarkusApplicationPropertiesContext extends TemplateContextType {
+
+    protected QuarkusApplicationPropertiesContext() {
+        super("QUARKUS_APPLICATION_PROPERTIES", "Quarkus properties");
+    }
+
+    @Override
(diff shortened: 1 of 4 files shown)


angelozerr:
@sdaschner I close this PR for snippets support, the main idea is that it the language server which manages snippets (like MicroProfile LS / Qute LS, etc) and not the LSP client (Intellij in this case).

Managing snippet on language server side gives the capability:

 * to show snippet not anywhere (according the AST node manages on language server)
 * any LSP client can benefit about those snippets, in other words snippets are consistent.

The write of snippet are written in JSON by using the LSP snippet syntax. See for instance for Quarkus 

 * snippets in java file https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.ls.ext/com.redhat.quarkus.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-java.json in Java file.
 * snippets in application.properties file https://github.com/redhat-developer/quarkus-ls/blob/master/quarkus.ls.ext/com.redhat.quarkus.ls/src/main/resources/com/redhat/quarkus/snippets/quarkus-properties.json

The only main problem with Intellij LSP client is that it doesn't support Templates variables which is very bad for snippets. I created the issue at https://github.com/redhat-developer/intellij-quarkus/issues/786

@sdaschner if you wish support your snippets, please contribute to Quarkus LS JSON snippets. Thanks.

I close this issue.

PR #75, opened 2020-04-21
Added highlighting to application properties

sdaschner:
Also added a missing icon (just a suggestion).


sdaschner:
Thanks for the review, updated.

PR #186, opened 2020-07-09
Add missing default cases in switch statements

paser4se:
Signed-off-by: PASER4SE <aarahman@ncsu.edu>

As mentioned in https://github.com/redhat-developer/intellij-quarkus/pull/183 added sign-off header ... will replace the existing two pull requests


paser4se:
According to CWE, not having a default case for switch statements in code is a security weakness. See https://cwe.mitre.org/data/definitions/478.html

paser4se:
> As indicated in the README, the whole intellij-community folder is an exact substract of the intellij-community open source project so we should not modify them. So can you remove this change from your PR

Which one ... HTTP or missing default?

jeffmaury:
Yes the http -> https switch as my comment was about `intellij-community` sub folder

paser4se:
Fixed the `http` issue.

PR #221, opened 2020-09-02
Selected extensions column is now scrollable

### LINKED ISSUE #213: Right column with "selected extensions" is not scrollable (when too many extensions are selected)
The right column with "selected extensions" in the new Quarkus project wizard is not scrollable. Additionally to that when a user is selecting more and more extensions the wizard window expands in height until it reaches its limits (until the wizard window is from the top to the bottom of the screen, and of course the window is not resizable anymore).

Snímek obrazovky 2020-08-25 v 21 40 28

jeffmaury:
Fixes #213

Signed-off-by: Jeff MAURY <jmaury@redhat.com>


zcervink (approved):
@jeffmaury I am approving this PR - I have tested it on localhost several times, the column is now scrollable without any issue. Thank you for the fix, it can really make the user experience better.

PR #268, opened 2020-10-21
Update the plugin runtime to version 1.9.0.Final

### LINKED ISSUE #267: Update the plugin runtime to version 1.9.0.Final
New version of the runtime is available, update it to version 1.9.0.Final.

zcervink:
Fixes #267

Signed-off-by: Zbynek Cervinka <zcervink@redhat.com>

Tested on localhost, new version of runtime is available and downloaded without any issue.

Please review @jeffmaury

### DIFF
--- a/projects/gradle/config-properties/gradle.properties
+++ b/projects/gradle/config-properties/gradle.properties
@@ -1,6 +1,6 @@
 #Gradle properties
 #Wed Jun 03 05:57:10 UTC 2020
-quarkusPluginVersion=1.8.3.Final
+quarkusPluginVersion=1.9.0.Final
 quarkusPlatformArtifactId=quarkus-universe-bom
 quarkusPlatformGroupId=io.quarkus
-quarkusPlatformVersion=1.8.3.Final
+quarkusPlatformVersion=1.9.0.Final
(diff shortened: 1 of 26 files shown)


zcervink:
@jeffmaury I have changed the title to **[WIP] ...**

I have found that the **Test testMicroprofileReactiveMessagingPropertiesTest** fails on all OSes. I was thinking about converting the PR to a draft, but I am not sure you will have access to it - it says "**People who are already subscribed will not be unsubscribed.**" - subscribed means that I have mentioned you on the PR?

zcervink:
@jeffmaury thank you very much for implementing the workflow for updating the runtime. This PR has already been created, so I think we should finish it and let the workflow generate next PRs.

### 3 checks were not successful, there were probably some changes between 1.8.x and 1.9.x versions of the runtime:
```
Test testMicroprofileReactiveMessagingPropertiesTest[0: with Gradle-6.0] FAILED (48.6s)

java.lang.AssertionError: mp.messaging.connector.smallrye-kafka.cloud-events=2 expected:<0> but was:<1>
```
I think this PR is OK, the tests should be updated, but I am not sure (probably the tests should be updated as a part o this PR, just change the expected value to 1).

jeffmaury:
There is no simple way to change the expected count because the check is verifying that properties are not duplicated.
So I suggest you just comment the check in the test so that build passes ok

zcervink:
> There is no simple way to change the expected count because the check is verifying that properties are not duplicated.
> So I suggest you just comment the check in the test so that build passes ok

I have investigated the issue and I have commented one line of code with an assert. I have also added a comment with description above the commented line so it is clear why it is commented.

PR #280, opened 2020-10-29
Switch Quarkus runtime to 1.9.1.Final

jbosstoolsci:


zcervink:
@jeffmaury The Signed-off-by is missing and the checks are failing due to the issue we have discussed in "update to 1.9.0.Final" PR. Despite the fact, that newer runtime than 1.9.0 is out, I suggest to merge the https://github.com/redhat-developer/intellij-quarkus/pull/268 first. There is the commented line of code that used to cause the failure of the tests/checks. After that update the workflow to add the Signed-off-by and generate new PR -> all checks should be green.

PR #296, opened 2020-11-12
Switch Quarkus runtime to 1.10.0.CR1

jbosstoolsci:
Signed-off-by: jbosstoolsci Bot <jbosstoolsci@jboss.org>


zcervink:
@jeffmaury maybe we should update the workflow to skip the CR* versions and wait for the Final one. Or just leave the workflow as it is and just delete the PR (I think we are not planning to update to CR* (it seems like it is a beta/not final version).

jeffmaury:
See #297

PR #302, opened 2020-11-20
Add steps to run UI tests to the README.md

zcervink:
Fixes 287

Signed-off-by: Zbynek Cervinka <zcervink@redhat.com>

@jeffmaury Please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,14 @@ This script is ```pulltest.sh```
 If the version of the IntelliJ SDK used to build is changed (see gradle.properties), you must
 update the branch in ```pulltest.sh``` and run the script again.
 
+#### UI testing
+
+You can perform UI testing by running the following command:
+
+```sh
+./gradlew clean runIdeForUiTests -PideaVersion=2020.1& ./gradlew integrationTest 
+```
+
 ## Feedback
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,6 +195,14 @@ This script is ```pulltest.sh```
 If the version of the IntelliJ SDK used to build is changed (see gradle.properties), you must
 update the branch in ```pulltest.sh``` and run the script again.
 
+#### UI testing
+
+You can perform UI testing by running the following command:
+
+```sh
+./gradlew clean runIdeForUiTests -PideaVersion=2020.1& ./gradlew integrationTest
```
jeffmaury, line 203: Are you sure of the syntax ? Aren't we missing && between the 2 commands

zcervink: Yes, I am sure, I have run the tests this way many times (there is probably no other way to run the tests as one line cmd). && are used as a connection between 2 cmds where the first cmd runs and after it will finish successfully the second cmd will run (in case the first cmd do not finish successfully the second cmd will never run). We have different situation - I use & to run first cmd in background, second cmd will run in parallel. There is no other way than run the 2 cmds (processes) in parallel. Running them one after the other will not work. When & is used, there is no need for anything more such semicolon etc.

PR #417, opened 2021-06-30
feat: Connect shared library for UI tests and remove unnecessary code which has been moved to the shared library

### LINKED ISSUE #408: Connect shared library for UI tests to IntelliJ Quarkus plugin project
- extend the build.gradle file to contain dependency to the shared library (make the shared library connected and accessible in the intellij-quarkus project)
- remove unnecessary gradle configuration (dependency on Remote-Robot framework and other configuration that is packed in the shared library)
- remove unnecessary Java code (fixtures and utils that has been moved to the shared library)

zcervink:
Fixes #408

Please review @jrichter1 - I think we are ready to make this PR - tested on localhost, cooperation with shared library works.


zcervink:
@jeffmaury Ready for review. I have rebased the PR and I have updated the version of the shared library to the newest version (released today with some fixes and UI test stability improvements). The tests looks now quite stable, sometime the rerun is needed due to some instability (I am working on that).

zcervink:
@jeffmaury Years updated in new files (+ I hope we are doing still the strategy - after review/comments new commits should be added instead of amending old commits - additional changes are more clear and visible. If not I can squash everythingh into one commit.)

Well - having one workflow file instead of 3 will indeed be more maintainable. But github does not allow you to rerun integration tests on only one single operation system when you have the 3 OSes in one workflow file. **And this is big issue for UI tests** - @jrichter1 suggested this approach of splitting the itests into more files several time in the past and he is also using this approach. In general integration tests are not stable as other builds/Junit tests, rerunning will be probably needed once in a while. Reruning successful testrun can result into failing test. I really suggest to do an exception for UI tests and keep it separated into 3 files. Increasing number of integration tests can lead to situation that you do not get all ticks green even after several reruns.

zcervink:
@jeffmaury After yesterday's discussion I kept the 3 separated workflows for integration tests. Do you have anything else?

build.gradle
```diff
@@ -129,10 +128,7 @@ dependencies {
     //required by lsp4j as the version from IJ is incompatible
     compile 'com.google.code.gson:gson:2.8.9'
     compile 'com.vladsch.flexmark:flexmark:0.50.50'
-    integrationTestImplementation("com.intellij.remoterobot:remote-robot:$remoteRobotVersion")
-    integrationTestImplementation("com.intellij.remoterobot:remote-fixtures:$fixturesVersion")
-    integrationTestImplementation 'org.junit.jupiter:junit-jupiter-api:5.3.1'
-    integrationTestRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.3.1'
+    compile 'com.redhat.devtools.intellij:intellij-common-ui-test-library:0.0.8'
```
jeffmaury, line 131: If it's in compile scope then it's going to finish in the plugin ZIP file ?

zcervink: Yes, it will, I have tested that on localhost. Do you have an idea/suggestion on improving that? Another way that works for UI tests+lib connection but does not include that in the plugin ZIp file?

zcervink: I see the **testCompile** which could probably work. After `./gradlew buildPlugin` the library is not in the plugin ZIP file. I have never used that, do you think I should replace that?

jeffmaury: Yes I see no reason to have that lib in the production ZIP file

zcervink:
@jeffmaury Fixed, all checks green.

jeffmaury (approved):
LGTM

Issue #472, opened 2021-10-14
Exception when using env. variables for the http port property


tlvlp:
The application.properties parser assumes that http port will only contain a parseable integer and throws the following exception.

application.properties
```properties
quarkus.http.port=${PORT:8081}
```

Exception:

```log
Error while converting '${PORT:8081}' as Integer for key 'quarkus.http.port'

java.lang.NumberFormatException: For input string: "${PORT:8081}"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:638)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.core.project.AbstractConfigSource.getPropertyAsInt(AbstractConfigSource.java:149)
	at com.redhat.devtools.intellij.lsp4mp4ij.psi.core.project.PsiMicroProfileProject.getPropertyAsInteger(PsiMicroProfileProject.java:81)
	at com.redhat.microprofile.psi.internal.quarkus.jaxrs.java.QuarkusJaxRsCodeLensParticipant.beginCodeLens(QuarkusJaxRsCodeLensParticipant.java:44)
	at com.redhat.devtools.intellij.lsp4mp4ij.psi.core.PropertiesManagerForJava.lambda$collectCodeLens$3(PropertiesManagerForJava.java:144)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at com.redhat.devtools.intellij.lsp4mp4ij.psi.core.PropertiesManagerForJava.collectCodeLens(PropertiesManagerForJava.java:144)
	at com.redhat.devtools.intellij.lsp4mp4ij.psi.core.PropertiesManagerForJava.lambda$codeLens$1(PropertiesManagerForJava.java:122)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:841)
	at com.redhat.devtools.intellij.lsp4mp4ij.psi.core.PropertiesManagerForJava.codeLens(PropertiesManagerForJava.java:115)
	at com.redhat.devtools.intellij.quarkus.lsp.QuarkusLanguageClient.lambda$getJavaCodelens$11(QuarkusLanguageClient.java:154)
[...]

PR #474, opened 2021-10-15
fix: Update http port parsing with possible env variables (#472)

tlvlp:


jeffmaury:
Fixed by e7db52191364268252c46677b206f67898ed2574

Issue #755, opened 2023-03-24
1.16 made JetBrain Gradle plugin mandatory


jesnil69:
```
IntelliJ IDEA 2022.3.3 (Community Edition)
Build #IC-223.8836.41, built on March 10, 2023
Runtime version: 17.0.6+1-b653.34 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Windows 10 10.0
GC: G1 Young Generation, G1 Old Generation
Memory: 2018M
Cores: 12
Non-Bundled Plugins:
    com.redhat.devtools.intellij.telemetry (0.0.3.33)
    com.redhat.devtools.intellij.quarkus (1.16.0.329)

Kotlin: 223-1.8.0-release-345-IJ8836.41
```

IntelliJ repeatedly shows the following exception while working in both Quarkus and other types of projects when the Quarkus plug-in is enabled:

```
Internal error: java.lang.NoClassDefFoundError: org/jetbrains/plugins/gradle/util/GradleConstants

java.util.concurrent.CompletionException: java.lang.NoClassDefFoundError: org/jetbrains/plugins/gradle/util/GradleConstants
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:708)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
	at com.redhat.devtools.intellij.quarkus.lsp4ij.IndexAwareLanguageClient.runTask(IndexAwareLanguageClient.java:101)
	at com.redhat.devtools.intellij.quarkus.lsp4ij.IndexAwareLanguageClient$1.run(IndexAwareLanguageClient.java:76)
	at com.intellij.openapi.progress.impl.CoreProgressManager.startTask(CoreProgressManager.java:429)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.startTask(ProgressManagerImpl.java:114)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcessWithProgressInCurrentThread$11(CoreProgressManager.java:557)
[...]

PR #756, opened 2023-03-24
Make Gradle dependency optional again

jesnil69:


src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/core/providers/GradleProjectLabelProvider.java
```diff
@@ -31,11 +33,17 @@
 public class GradleProjectLabelProvider implements IProjectLabelProvider {
 
 	public static final String GRADLE_LABEL = "gradle";
+	private static final PluginManager instance = PluginManager.getInstance();
+	private static final PluginId gradlePluginId = PluginId.findId("com.intellij.gradle");
```
adietish, line 37: These are not constant expressions, I'd thus preferably not declare them as static. I'd rather use an instance var any maybe do lazy loading.

jesnil69: Hi @adietish, thanks for the review.
Actually, this type of conditional qualification is not needed at all. The correct solution is to not make these Gradle dependent providers mandatory, only load them when Gradle plug-in is enabled.
See #755 for the correct and simple solution.

PR #767, opened 2023-04-04
chore: simplified validation workflow

sbouchet:


jeffmaury:
> @sbouchet: the build complains that we should bump the `since-version`:
> 
> ```
> [gradle-intellij-plugin :verifyPluginConfiguration] The following plugin configuration issues were found:
> - The 'since-build' property is lower than the target IntelliJ Platform major version: 203 < 212.
> ```

Sounds logical to me and the verifier is really dumb in that case

sbouchet:
> > @sbouchet: the build complains that we should bump the `since-version`:
> > ```
> > [gradle-intellij-plugin :verifyPluginConfiguration] The following plugin configuration issues were found:
> > - The 'since-build' property is lower than the target IntelliJ Platform major version: 203 < 212.
> > ```
> 
> Sounds logical to me and the verifier is really dumb in that case

yes, pluginVerifier prints the message for all versions that are not the default one in properties. see for example the intellij-openshift-connector checks

adietish (approved):

PR #778, opened 2023-04-11
fix: Synchronize correctly the IJ document with the LSP server document

angelozerr:
Synchronize correctly the IJ document with the LSP server document.

When you type fastly some content inside the editor content (ex : you open an html file to consume the Qute LS and you type a lot of spaces (let's press space key to generate 20 characters for instance), the IJ document is desynchronized with the LSP server document.

One this desynchronization occurs, the Qute support is unusable (ex : the validation report bad error ranges). 

This PR fixes this problem by avoiding lossing document change event.

Here a demo with the problem in master:

IJDocumentSyncBug

Here the same demo with this PR which should fix this issue: 

IJDocumentSyncFix

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/DocumentContentSynchronizer.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/DocumentContentSynchronizer.java
@@ -25,23 +25,27 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
-import java.io.File;
 import java.net.URI;
-import java.util.Collections;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


angelozerr:
@adietish @jeffmaury @sbouchet I think this PR is very important because when you type fast, the editor is desynchronized with the language server and it is unusable.

adietish:
I can't replicate the issue on MacOS :( So I can't tell that it fixes something. Looking at the code though the change makes sense to me.

adietish:
I can replicate other problems: the for-suggestion does not appear and I get weird highlighting in a html page when quickly cutting & pasting:
image
With the patch the highlighting errors dont appear any more.

adietish (approved):

jeffmaury (changes requested):
Please rewrite commit message and PR title to follow conventional commit messages

PR #792, opened 2023-04-13
Diagnostics are out of sync

### LINKED ISSUE #692: Diagnostics are out of sync
The LSP4MP diagnostics info seems to lag behind the user changes, showing diagnostic info from a previous state of the document before the change.
Here, a Bulkhead value of `-1` does get underlined in red, but there is no hover info: 
image

Adding a second digit, the diagnostic info updates to show the previous error:
image

When quickly entering a few more characters, a seemingly random number of changes get processed in the diagnostic update, but never seems to be fully up-to-date with the changes. In both the following cases, I started with `-12` and quickly added `1234` in one quick stroke.
image
image

When I highlight the entire value and fix it to something appropriate, the diagnostic for the old value still remains, while also keeping the red underline for the same number of characters in the previous error (you can see it extend to part of `public` a couple lines down):
image

Environment info: 
Quarkus tools: 1.14.0.292
IntelliJ 2022.2.3
OS: Reproduced on both macOS 13.0.1 and Windows 10
Plugins downloaded: 
image

angelozerr:
Diagnostics are out of sync

Fixes #692

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/PropertiesManagerForJava.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/PropertiesManagerForJava.java
@@ -474,12 +474,12 @@ public class PropertiesManagerForJava {
      *
      * @param unresolved the CodeAction to resolve
      * @param utils      the utilities class
-     * @param monitor    the monitor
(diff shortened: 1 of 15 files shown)


angelozerr:
I close this PR since it was a POC which fixes several issues in same time. It will be very hard to review it, so I close it to create fixes with several PRs.

PR #793, opened 2023-04-13
fix: Codelens URL JAX-RS slows down the Java editor

### LINKED ISSUE #790: Codelens URL JAX-RS slows down the Java editor
When Java class is annotated with @Path in class, the Java editor is very slow. After some investigation, the problem comes from the URL codelens with https://github.com/redhat-developer/intellij-quarkus/blob/02d2869c4ebc89ebead5863fe2b3a372c2c28b3d/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/jaxrs/java/JaxRsCodeLensParticipant.java#L139 which checks if server is available. If you update the method by returning true, th eperformance are very good.

To reproduce the problem create a Foo class like this:

```java
package org.acme.rest.client;

import javax.ws.rs.Path;

@Path("")
public class Foo {

    
}

When you comment @Path annotation, the codelens URL is not computed and it is very fast, if you have @Path annotation it is very slow.

See the following demo:

```
CodeLensURLIJSlow

@jeffmaury have you some idea how we could fix this issue? The easy fix is to remove this check of server available, but perhaps a better fix should be done in https://github.com/redhat-developer/intellij-quarkus/blob/02d2869c4ebc89ebead5863fe2b3a372c2c28b3d/src/main/java/com/redhat/devtools/intellij/quarkus/lsp/QuarkusLanguageClient.java#L147 by improving runAsBackground?

angelozerr:
Codelens URL JAX-RS slows down the Java editor

Fixes #790

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/jaxrs/java/JaxRsCodeLensParticipant.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/jaxrs/java/JaxRsCodeLensParticipant.java
@@ -136,11 +136,17 @@ public class JaxRsCodeLensParticipant implements IJavaCodeLensParticipant {
 	}
 
 	private static boolean isServerAvailable(String host, int port, int timeout) {
-		try (Socket socket = new Socket()) {
+		/*try (Socket socket = new Socket()) {
 			socket.connect(new InetSocketAddress(host, port), timeout);
 			return true;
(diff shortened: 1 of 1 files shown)


angelozerr:
This PR removes the check of server available to avoid having freeze of the Java Editor as soon as user type something.

See demo at https://github.com/redhat-developer/intellij-quarkus/issues/790#issue-1664094920 how to reproduce the freeze.

adietish (approved):

Issue #799, opened 2023-04-13
@RestClient is not suggested if I undo a first insertion. If waiting it appears again.


adietish:
follows up on #781 

**Steps:**
1. EXEC: have `git@github.com:adietish/code-with-quarkus.git` cloned and imported into IJ
2. EXEC: open `ExtenstionsResource.java` in an editor
3. EXEC: put cursor into `extensionsService` and activate the suggestions (press Alt + Return)
```java
    @Inject
    ExtensionsService extensionsService;
```
4. ASSERT: "Insert @RestClient" is suggested
<img width="538" alt="image" src="https://user-images.githubusercontent.com/25126/228841622-d9cd51af-ea7b-4a56-bd70-329bd7ed82ce.png">

5. EXEC: select it
6. ASSERT: `@RestClient` is inserted above `@Inject`
7. EXEC: Undo
8. EXEC: activate the suggestions again an try to find "Insert @RestClient"

</br>

**Result:**
"Insert @RestClient" is not present in the suggestions. If you wait for maybe 20s it will be present again.
<img width="499" alt="image" src="https://user-images.githubusercontent.com/25126/231783058-3f57b9ad-6eef-4bbf-9a36-4fa7cc0af0b3.png">

angelozerr:
Not really :)

evie-lau:
This may possibly be related to the diagnostic sync issue #692

If you make a document change (adding or deleting a harmless whitespace), does it update instantly?

angelozerr:
@evie-lau @adietish I'm rewriting diagnotic and quick fix in draft PR https://github.com/redhat-developer/intellij-quarkus/pull/792

For the moment, you need to click on the error (carret change) to compute the quick fixes.

But it should fix this issue (if you click where there is the error).

@evie-lau I noticed that you maintain a copy/paste of our repository:

 * https://github.com/OpenLiberty/liberty-tools-intellij/tree/main/src/main/java/io/openliberty/tools/intellij/lsp4mp
 * https://github.com/OpenLiberty/liberty-tools-intellij/tree/main/src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi

IMHO I think we should have several plugins:

 * an LSP plugin (like vscode)
 * a MicroProfile plugin, that IBM could be based on (like vscode-microprofile)
 * a Quarkus plugin  (like vscode-quarkus)

@jeffmaury what do you think about this idea?

angelozerr:
[...]

PR #805, opened 2023-04-18
fix: Diagnostics are out of sync

### LINKED ISSUE #692: Diagnostics are out of sync

angelozerr:
fix: Diagnostics are out of sync

Fixes #692


adietish (approved):
tested it with an annotation `Timeout(1000)` that I set it to an invalid negative value `Timeout(-1000)` and then back again. Validation worked correctly, indicated an invalid value and became valid again once the value was back to positive.

angelozerr:
> tested it with an annotation Timeout(1000) that I set it to an invalid negative value Timeout(-1000) and then back again. Validation worked correctly, indicated an invalid value and became valid again once the value was back to positive.

Nice! Thanks for having tested that.

And what about code action issue https://github.com/redhat-developer/intellij-quarkus/issues/799 ?

It should work better without blocking the validation process.

angelozerr:
Thanks so much @adietish for your review.

adietish:
@angelozerr:

> And what about code action issue #799 ?
> 
> It should work better without blocking the validation process.

didn't measure but it felt like there was less time needed before the suggestion was offered again (after undo of the initial insertion).

angelozerr:
> didn't measure but it felt like there was less time needed before the suggestion was offered again (after undo of the initial insertion).

Nice! I have another ideas how to improve again this problem.

angelozerr:
I think https://github.com/redhat-developer/intellij-quarkus/pull/809 should  improve again performance code action, please give me feedback.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/operations/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static void updateAnnotations(List<Annotation> existingAnnotations, AnnotationHolder holder) {
+        if (existingAnnotations == null || existingAnnotations.isEmpty()) {
+            return;
+        }
+        // Copy all existing annotations in the given annotation holder.
+        existingAnnotations.forEach(currentAnnotation -> {
+            // Copy range, severity, message information
+            TextRange range = new TextRange(currentAnnotation.getStartOffset() , currentAnnotation.getEndOffset());
+            Annotation newAnnotation = holder.createAnnotation(currentAnnotation.getSeverity(),  range, currentAnnotation.getMessage());
```
evie-lau, line 108: @angelozerr I tried mirroring these changes in liberty-tools-intellij. I'm getting this Exception whenever an annotation is getting created.
```
Caused by: com.intellij.util.IncorrectOperationException: 'AnnotationHolder.createAnnotation()' method (the call to which was found in class io.openliberty.tools.intellij.lsp4mp.lsp4ij.operations.diagnostics.LSPDiagnosticAnnotator) is slow, non-incremental and thus can cause unexpected behaviour (e.g. annoying blinking), is deprecated and will be removed soon. Please use `newAnnotation(...).create()` instead
```
I am running on IntelliJ 2023.1 with intellij gradle plugin 1.13.3, so possibly it is just deprecated in some of these newer versions.

angelozerr: Ok thanks for your feedback. You mean that it doesn't work or it is just a warning message?

angelozerr: I hav enot used this create method because it doesn't return the Annotation and I need to store it to update quick fix in async mode. I need to investigate more how to use create method.

angelozerr: OK thanks for your feedback. I have investigate how to use the non deprecated api with annotation and I have found a solution. 

I will ping you when I will create the PR

angelozerr: @evie-lau please see https://github.com/redhat-developer/intellij-quarkus/pull/819

PR #809, opened 2023-04-19
fix: CodeAction should be resolved only when user click on QuickFix

### LINKED ISSUE #785: CodeAction should be resolved only when user click on QuickFix
Qute / Quarkus LS provides some code action with resolve support, generally it is code action for Java file which is done with resolve.

It is very important to resolve code action on click on the quick fix and not when diagnostic is published. Today code action for Java file are resolved (we compute text edit for all code actions by using PsiJavaFile which can takes some times and are not very good for performance).

In other words the compute of text edit by using PsiJavaFile must be done on click of the quick fix and not when the editor is validated. If we have several code actions, today all code actions are resolved while the validation process, we compute them although user don't use it.

angelozerr:
fix: CodeAction should be resolved only when user click on QuickFix

Fixes #785

Signed-off-by: azerr <azerr@redhat.com>

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/PropertiesManagerForJava.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/PropertiesManagerForJava.java
@@ -474,12 +474,12 @@ public class PropertiesManagerForJava {
      *
      * @param unresolved the CodeAction to resolve
      * @param utils      the utilities class
-     * @param monitor    the monitor
      * @return the codeAction list according the given codeAction parameters.
      */
-    public CodeAction resolveCodeAction(CodeAction unresolved, IPsiUtils utils,
-                                        ProgressIndicator monitor) {
-        return codeActionHandler.resolveCodeAction(unresolved, utils);
+    public CodeAction resolveCodeAction(CodeAction unresolved, IPsiUtils utils) {
+        return ApplicationManager.getApplication().runReadAction((Computable<CodeAction>) () -> {
+            return codeActionHandler.resolveCodeAction(unresolved, utils);
+        });
     }
 
 }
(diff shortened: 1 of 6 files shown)


angelozerr:
This PR proides the capability to support LSP resolve code action. The main idea is if code action can support resolve, the code action will doesn't compute the text edit which can be very expansive for Java quick fix. To test this PR, you can test it with the `Insert @RestClient ` quick fix.

Performance should be better (quick fix should be updated quickly) and CPU usage should be better too (if you do a lot of Undo / Redo).

To check this PR, I suggest that you set a breakpoint when code action is collected https://github.com/redhat-developer/intellij-quarkus/blob/5230c8d2750df362faf68bad5ce9a1eb5d4cdf59/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPVirtualFileWrapper.java#L165

and takes a Java file which requires @RestClient. If you type in the editor something,it should stop at this breakpoint:

image

and the edit of CodeAction should be to null.

If you click on CodeAction, the edit should be resolved at this step (the @restClient must be inserted in the editor).

adietish:
> This PR proides the capability to support LSP resolve code action. The main idea is if code action can support resolve, the code action will doesn't compute the text edit which can be very expansive for Java quick fix. To test this PR, you can test it with the `Insert @RestClient ` quick fix.
> 
> Performance should be better (quick fix should be updated quickly) and CPU usage should be better too (if you do a lot of Undo / Redo).
> 
> To check this PR, I suggest that you set a breakpoint when code action is collected
> 
> https://github.com/redhat-developer/intellij-quarkus/blob/5230c8d2750df362faf68bad5ce9a1eb5d4cdf59/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPVirtualFileWrapper.java#L165
> 
> and takes a Java file which requires @RestClient. If you type in the editor something,it should stop at this breakpoint:
> 
> image
> 
> and the edit of CodeAction should be to null.
> 
> If you click on CodeAction, the edit should be resolved at this step (the @restClient must be inserted in the editor).

I can confirm this in terms of values at the brk point but also in performance. Things are much smoother. The other additional improvement here is that #799 feels fixed. There's no delay after an undo for the @RestClient to be suggested.
Giving it a strong +1.

adietish (approved):
a lot of improvement here. +1

angelozerr:
> a lot of improvement here. +1

Thanks so much for your review @adietish 

I have another ideas to continue the performance improvement. I will create another PRs.

Issue #807, opened 2023-04-19
CodeAction must be collected on hover or caret changed


angelozerr:
follows up on #805 

Today code actions are collected on validation step by looping each diagnostics and compute code action for each diagnostics.

It is bad:

 * for performance, instead of collecting code action for all diagnostics, we need to collect only code action of the curent caret cursor location
 * it exists some code actions (not in MP / Qute LS) which could be not linked to a diagnostic like refactor code actions, we don't takes care of this kind of code action.

To fix those 2 problems, we need to collect code action:

 * when user hover an error.
 * when caret location changed (without text document changed), in other words when user use keybord to move only cursor or click with mouse.

angelozerr:
After some discussion with @fbricon it is a bad idea to update quickfix on caret and mouse over because it breaks IJ features like QuickFix popup on problem view.

I abandonned this PR and I replace it with https://github.com/redhat-developer/intellij-quarkus/pull/819

PR #816, opened 2023-04-25
feat: CodeAction must be collected on hover or caret changed (#807)

angelozerr:
feat: CodeAction must be collected on hover or caret changed (#807)

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
@@ -1,0 +1,90 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
+package com.redhat.devtools.intellij.quarkus.lsp4ij;
+
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.util.Key;
+import com.redhat.devtools.intellij.quarkus.lsp4ij.operations.codeactions.LSPCodeActionUpdater;
+
+/**
+ * LSP editor wrapper used to:
+ *
+ * <ul>
+ *     <li>collect LSP code action only when an LSP diagnostic is hovered ot when caret cursor changed.</li>
+ * </ul>
+ */
+public class LSPEditorWrapper {
+
+    private static final Key<LSPEditorWrapper> KEY = Key.create(LSPEditorWrapper.class.getName());
+
+    private final Editor editor;
+
+    private final LSPCodeActionUpdater codeActionUpdater;
+
+    private LSPEditorWrapper(Editor editor) {
+        this.editor = editor;
+        this.codeActionUpdater = new LSPCodeActionUpdater(editor);
+    }
+
+
+    /**
+     * Returns the Intellij editor instance which is wrapped.
+     *
+     * @return the Intellij editor instance which is wrapped.
+     */
+    public Editor getEditor() {
+        return editor;
+    }
+
+    /**
+     * Returns true if the given Intellij editor is wrapped and false otherwise.
+     * @param editor the Intellij editor instance.
+     * @return true if the given Intellij editor is wrapped and false otherwise.
+     */
+    public static boolean hasWrapper(Editor editor) {
+        return editor.getUserData(KEY) != null;
+    }
+
+    /**
+     * Dispose the LSP editor wrapper.
+     */
+    public void dispose() {
+        this.codeActionUpdater.dispose();
+        editor.putUserData(KEY, null);
+    }
+
+    /**
+     * Returns the LSP editor wrapper from the given Intellij editor.
+     * @param editor the Intellij editor instance
+     * @return the LSP editor wrapper from the given Intellij editor.
+     */
+    public static LSPEditorWrapper getLSPEditorWrapper(Editor editor) {
+        LSPEditorWrapper wrapper = editor.getUserData(KEY);
+        if (wrapper != null) {
+            return wrapper;
+        }
+        return getLSPEditorWrapperSync(editor);
+    }
+
+    private static synchronized LSPEditorWrapper getLSPEditorWrapperSync(Editor editor) {
+        LSPEditorWrapper wrapper = editor.getUserData(KEY);
+        if (wrapper != null) {
+            return wrapper;
+        }
+        wrapper = new LSPEditorWrapper(editor);
+        editor.putUserData(KEY, wrapper);
+        return wrapper;
+    }
+
+}
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPVirtualFileWrapper.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPVirtualFileWrapper.java
@@ -10,7 +10,6 @@
  *
  * Contributors:
  *     Red Hat Inc. - initial API and implementation
- *     Some piece of code has been inspired by https://github.com/ballerina-platform/lsp4intellij/blob/master/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
  *******************************************************************************/
 package com.redhat.devtools.intellij.quarkus.lsp4ij;
 
@@ -47,8 +46,6 @@ public class LSPVirtualFileWrapper {
     private final VirtualFile file;
 
     private final Map<String /* language server id */, List<Diagnostic> /* current diagnostics */> diagnosticsPerServer;
-
-    private List<Annotation> annotations;
 
     private boolean diagnosticDirty;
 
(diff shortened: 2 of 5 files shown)


src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+    private boolean documentChanged;
+
+    private LSPCodeActionResult codeActionResult;
+
+    private LSPEditorWrapper(Editor editor) {
+        this.editor = editor;
+        this.editorMouseMotionListener = new EditorMouseMotionListener() {
+            @Override
+            public void mouseMoved(@NotNull EditorMouseEvent e) {
+                Document document = editor.getDocument();
```
adietish, line 55: as far as I remember, retrieving the document requires a ReadOperation?

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+                                try {
+                                    updateQuickFixAt(offset,startOffset,endOffset).get();
+                                } catch (InterruptedException interruptedException) {
+                                    interruptedException.printStackTrace();
+                                } catch (ExecutionException executionException) {
+                                    executionException.printStackTrace();
+                                }
+                            }
+                        }
+                    }
```
adietish, line 74: nitpicking: those lines could be separated into their own method improving readability

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+                if (documentChanged) {
+                    documentChanged = false;
+                    return;
+                }
+                int offset = editor.getCaretModel().getCurrentCaret().getOffset();
+                VirtualFile file = LSPIJUtils.getFile(editor.getDocument());
+                if (file != null) {
+                    updateQuickFixAt(offset, -1,-1);
+                }
+            }
```
adietish, line 102: nitpicking: I believe that code readability would gain a lot if the 3 anonymous callbacks above would be separated into their own methods or inner classes

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+                List<Diagnostic> currentDiagnostics = LSPVirtualFileWrapper.getLSPVirtualFileWrapper(file).getAllDiagnostics();
+                if (currentDiagnostics != null && !currentDiagnostics.isEmpty()) {
+                    int offset = getTargetOffset(e);
+                    if (offset != -1) {
+                        for (Diagnostic diagnostic : currentDiagnostics) {
+                            int startOffset = LSPIJUtils.toOffset(diagnostic.getRange().getStart(), document);
+                            int endOffset = LSPIJUtils.toOffset(diagnostic.getRange().getEnd(), document);
+                            if (offset >= startOffset && offset <= endOffset) {
+                                try {
+                                    updateQuickFixAt(offset,startOffset,endOffset).get();
```
adietish, line 66: `updateQuickFixAt` can return `null`. Should `null`-checked before calling `get()` on the `CompletableFuture`

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+                                } catch (InterruptedException interruptedException) {
+                                    interruptedException.printStackTrace();
```
adietish, line 68: Shouldn't we better LOG this error?

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+                                } catch (ExecutionException executionException) {
+                                    executionException.printStackTrace();
```
adietish, line 70: Shouldn't we better LOG this error?

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+        if (!currentDiagnostics.isEmpty()) {
+            List<Diagnostic> diagnostics = new ArrayList<>(currentDiagnostics);
+            diagnostics.forEach(diagnostic -> {
+                int startOffset = LSPIJUtils.toOffset(diagnostic.getRange().getStart(), document);
+                int endOffset = LSPIJUtils.toOffset(diagnostic.getRange().getEnd(), document);
+                if (offset >= startOffset && offset <= endOffset) {
+                    diagnosticContext.add(diagnostic);
+                }
+            });
+        }
```
adietish, line 156: nitpicking: l146-156 could be separated into their own method ex. `createDiagnosticContext` and improve readability.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+        Project project = LSPIJUtils.getProject(file).getProject();
+        return LanguageServiceAccessor.getInstance(project)
+                .getLanguageServers(document, capabilities -> capabilities.getCodeActionProvider() != null)
+                .thenComposeAsync(languageServers -> CompletableFuture.allOf(languageServers.stream()
+                        .map(languageServer -> languageServer.getTextDocumentService().codeAction(params)
+                                .thenAcceptAsync(codeActions -> {
+                                    // textDocument/codeAction may return null
+                                    if (codeActions != null) {
+                                        this.codeActionResult = new LSPCodeActionResult(context, codeActions, languageServer);
+                                        updateErrorAnnotations(document, project, editor, offset, start, end);
```
adietish, line 170: why not pass `this.codeActionResult` as a parameter to `updateErrorAnnotations`?
I see that the instance var is required elsewhere but I believe that in this particular call it would be better to pass them to the method directly.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+                    LanguageServer languageServer = codeActionResult.getLanguageServer();
+                    List<Either<Command, CodeAction>> codeActions = codeActionResult.getCodeActions();
+
+                    for (Either<Command, CodeAction> commandOrCodeAction : codeActions) {
+                        LSPCodeActionIntentionAction quickFix = new LSPCodeActionIntentionAction(commandOrCodeAction,
+                                languageServer);
+                        TextRange textRange = new TextRange(start != -1 ? start : offset, end != -1 ? end : offset);
+                        HighlightDisplayKey k = HighlightDisplayKey.find("Annotator");
+                        info.registerFix(quickFix, null, HighlightDisplayKey.getDisplayNameByKey(k), textRange, k);
+                    }
```
adietish, line 201: nitpicking: l195-201 could be extracted to a method `registerQuickFixes`

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,246 @@
...
+
+    public static LSPEditorWrapper getLSPEditorWrapper(Editor editor) {
+        LSPEditorWrapper wrapper = editor.getUserData(KEY);
+        if (wrapper != null) {
+            return wrapper;
+        }
+        return getLSPEditorWrapperSync(editor);
+    }
+
+    private static synchronized LSPEditorWrapper getLSPEditorWrapperSync(Editor editor) {
```
adietish, line 236: why is this method called `getLSPEditorWrapperSync`? All I see in differences when compared to `getLSPEditorWrapper` is that it would create the wrapper if it doesn't exist. Shouldn't it be called `getOrCreateLSPEditorWrapper`?

adietish:
I believe that there is an edge case that is not handled correctly yet:

**Steps:**
1. ASSERT: Have the following class:
```
package org.acme.rest.client;

import org.eclipse.microprofile.faulttolerance.Timeout;
import org.eclipse.microprofile.rest.client.inject.RestClient;
import org.jboss.resteasy.annotations.jaxrs.PathParam;

import javax.inject.Inject;
import javax.ws.rs.GET;
import javax.ws.rs.Path;
import java.util.Collections;
import java.util.Set;

@RestClient
@Path("/extension")
public class ExtensionsResource {

    @Inject
    ExtensionsService extensionsService;

    @GET
    @Path("/id/{id}")
    @Timeout(1)
    public Set<Extension> id(@PathParam String id) {
        return extensionsService.getById(id);
    }
}
```
2. EXEC: position your cursor to `extensionsService` and hit "Alt + Enter"
3. ASSERT: suggestion "Add @RestClient" is present in list of possible actions.
4. EXEC: select "Insert @RestClient" which causes it to be inserted above `extensionsService` and then hit Ctrl + Z (undo).
5. ASSERT: "@RestClient" disappears from source
6. EXEC: hit Alt + Enter (without moving the mouse nor the cursor"

**Result:**
"Insert @RestClient" is not suggested.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPEditorWrapper.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.util.Key;
+import com.redhat.devtools.intellij.quarkus.lsp4ij.operations.codeactions.LSPCodeActionUpdater;
+
+/**
+ * LSP editor wrapper used to:
+ *
+ * <ul>
+ *     <li>collect LSP code action only when an LSP diagnostic is hovered ot when caret cursor changed.</li>
```
fbricon, line 24: or when caret position changed

angelozerr:
After some discussion with @fbricon it is a bad idea to update quickfix on caret and mouse over because it breaks IJ features like QuickFix popup on problem view.

I abandonned this PR and I replace it with https://github.com/redhat-developer/intellij-quarkus/pull/819

Issue #806, opened 2023-04-18
Align with lsp4mp 0.7.1


datho7561:
This is an epic issue for tracking what needs to be done in order to have all the features from lsp4mp 0.7.1 in intellij-quarkus. I've included a list of all the commits that modify the JDT component of lsp4mp.

TODO:
- [ ] [Documentation on hover for properties in properties files](https://github.com/eclipse/lsp4mp/commit/2d6a264821fde2aaedb994625a88a4ae760c5f99) (large PR)
- [ ] [Fix definition for property value of user-defined class](https://github.com/eclipse/lsp4mp/commit/4c423f2a15c711d975eff836364e13cabec779cc) (depends on previous feature)
- [x] [~~Ensure config_ordinal only appears in MicroProfile projects~~](https://github.com/eclipse/lsp4mp/commit/caf91c68e7f2fe27072f1f947e3c5476f3289d8f)
- [x] ~~[Fix diagnostics data deserialization](https://github.com/eclipse/lsp4mp/commit/98a5f9988fbbf6c6ac8dea943c125aef611dd931)~~ only applies to JDT.LS delegate command handler
- [x] ~~[Prevent NPE when there are multiple definitions of a config property](https://github.com/eclipse/lsp4mp/commit/1a6462346822004ca8995bd6732434aba8b3b5c4#diff-b401157eef2dcd23fd582d334b3cf9a1978ba95f1d9d71525beb41cc494ac302)~~ only concerns LS bits
- [ ] [Show/hide Java snippets based on cursor location](https://github.com/eclipse/lsp4mp/commit/1ec7a763ca6e18e83c0feea37427812f9dcd2cb3) and [Fix context-aware snippets in projects that use Project Lombok](https://github.com/eclipse/lsp4mp/commit/e32232e52b9d9d6a0b3b3b7957939e9091b60721) (large PR)
[...]

Issue #803, opened 2023-04-14
Add support for Quarkus 3 and the jakarta namespace


fbricon:
This requires updating lsp4mp and quarkus-ls to the latest releases

fbricon:
Done in 1.18

PR #817, opened 2023-04-25
feat: Support jakarta namespace with Inject annotation

fbricon:
Port of https://github.com/eclipse/lsp4mp/commit/b416c804b9c2bc072ffa2010e728436ef899860b to IDEA
Part of #806 and #803

Fixes insertion of duplicate annotations

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/restclient/java/MicroProfileRestClientDiagnosticsParticipant.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/restclient/java/MicroProfileRestClientDiagnosticsParticipant.java
@@ -36,7 +36,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import static com.redhat.devtools.intellij.lsp4mp4ij.psi.core.MicroProfileConfigConstants.INJECT_ANNOTATION;
+import static com.redhat.devtools.intellij.lsp4mp4ij.psi.core.MicroProfileConfigConstants.INJECT_JAKARTA_ANNOTATION;
+import static com.redhat.devtools.intellij.lsp4mp4ij.psi.core.MicroProfileConfigConstants.INJECT_JAVAX_ANNOTATION;
 import static com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.restclient.MicroProfileRestClientConstants.REGISTER_REST_CLIENT_ANNOTATION;
 import static com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.restclient.MicroProfileRestClientConstants.REST_CLIENT_ANNOTATION;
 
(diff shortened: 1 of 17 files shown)


src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/restclient/java/MicroProfileRestClientDiagnosticsParticipant.java
```diff
@@ -118,7 +119,7 @@ private static void validateClassType(PsiClass classType, List<Diagnostic> diagn
 	private static void validateField(PsiField field, List<Diagnostic> diagnostics, JavaDiagnosticsContext context) {
 		String uri = context.getUri();
 		DocumentFormat documentFormat = context.getDocumentFormat();
-		boolean hasInjectAnnotation = AnnotationUtils.hasAnnotation(field, INJECT_ANNOTATION);
+		boolean hasInjectAnnotation = AnnotationUtils.hasAnnotation(field, INJECT_JAVAX_ANNOTATION) || AnnotationUtils.hasAnnotation(field, INJECT_JAKARTA_ANNOTATION);
```
angelozerr, line 122: I think we should be better to change the signature of hasAnnotation with `String... annotationNames` to write

`AnnotationUtils.hasAnnotation(field, INJECT_ANNOTATION, INJECT_JAKARTA_ANNOTATION);` because:

 * it is more readable
 * we avoid to loop twice the IJ Annotation

LSP4MP should adopt this signature too.

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/restclient/java/MicroProfileRestClientDiagnosticsParticipant.java
```diff
@@ -202,7 +203,7 @@ private static void validateInterfaceType(PsiClass interfaceType, List<Diagnosti
 			PsiElement o = PsiTreeUtil.getParentOfType(match.getElement(), PsiField.class);
 			if (o instanceof PsiField) {
 				PsiField field = (PsiField) o;
-				boolean hasInjectAnnotation = AnnotationUtils.hasAnnotation(field, INJECT_ANNOTATION);
+				boolean hasInjectAnnotation = AnnotationUtils.hasAnnotation(field, INJECT_JAVAX_ANNOTATION) || AnnotationUtils.hasAnnotation(field, INJECT_JAKARTA_ANNOTATION);
```
angelozerr, line 206: Same than https://github.com/redhat-developer/intellij-quarkus/pull/817/files#r1176382050

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/restclient/java/MicroProfileRestClientDiagnosticsParticipant.java
fbricon: when you have `hasAnnotation(String... annotation)`, it's not clear without reading the javadoc, whether this means has ALL or has ANY of the passed annotations. So I'd rename the method to `hasAnyAnnotation`

Now do you want to do that as part of this PR or can it be done in a separate one?

angelozerr:
I have played with this PR and it works like a charm and fix of import insert is a nice fix too.

Do you want to create an another PR with your idea with hasAnyAnnotation?

fbricon:
> I have played with this PR and it works like a charm and fix of import insert is a nice fix too.
> 
> Do you want to create an another PR with your idea with hasAnyAnnotation?

Yeah I'd rather prefer, as hasAnnotation is used in a lot of places

PR #819, opened 2023-04-26
Use newAnnotation API for diagnostics

angelozerr:
Use newAnnotation API for diagnostics and refresh code actions in background.


src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/operations/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -57,77 +54,44 @@ public LSPVirtualFileWrapper collectInformation(@NotNull PsiFile file, @NotNull
...
+        // Create Intellij Annotation from the given LSP diagnostic
         HighlightSeverity severity = toHighlightSeverity(diagnostic.getSeverity());
         TextRange range = new TextRange(start, end);
         String message = diagnostic.getMessage();
-        return holder.createAnnotation(severity, range, message);
+        AnnotationBuilder builder = holder.newAnnotation(severity, message)
+                .range(range);
+        // Register quick fixes if there are available
+        List<IntentionAction> fixes = diagnosticsForServer.getQuickFixesFor(diagnostic);
+        boolean codeActionsLoading = LSPDiagnosticsForServer.isCodeActionsLoading(fixes);
```
evie-lau, line 87: I found that these two lines should be moved above the AnnotationBuilder creation. I got some exceptions about `create()` not being called, maybe because there are some other actions in between initializing the builder and calling `create()`

angelozerr: I noticed this problem while I implementing this PR. If `diagnosticsForServer.getQuickFixesFor` throws some error the annotation is registered without create which causes your problem.

I move the line code as you suggested but now diagnosticsForServer.getQuickFixesFor should never thrown some exceptions.

angelozerr:
@fbricon @evie-lau @adietish the PR can be reviewed now (my code should be clean), please play with this PR and give me feedback. This PR brings:

 * the use of the AnnoationBuilder#newAnnotation API which is not deprecated
 * the load of code actions in background to avoid freezing the editor when validation is done
 * the cancel of code actions when they should not be used (ex : user type something, ols code action should be cancelled)
 * the reuse of code actions for a given diagnostic, if diagnostic has not changed from the previous validation.

adietish (approved):
Did extensive testing. 
We found a bug in `lsp4mpij` at `MicroProfileRestClientDiagnosticsParticipant.validateField()` where the `fieldType` was `null`, caused the language server to NPE and therefore not to suggest a required `@RestClient` annotation. 
This bug not being related to this PR I give it a +1 and will file an issue for the above bug.

angelozerr:
Thanks si mucg @adietish to have spent so many time to test my PR and give me feedback.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/operations/diagnostics/LSPDiagnosticAnnotator.java
```diff
@@ -57,77 +54,53 @@ public LSPVirtualFileWrapper collectInformation(@NotNull PsiFile file, @NotNull
...
-                if (diagnostic.getTags() != null && diagnostic.getTags().contains(DiagnosticTag.Deprecated)) {
-                    annotation.setHighlightType(ProblemHighlightType.LIKE_DEPRECATED);
-                }
-                annotations.add(annotation);
+        // Loop for language server which report diagnostics for the given file
+        diagnosticsPerServer.forEach(ds -> {
+            boolean codeActionsLoading = false;
+            // Loop for LSP diagnostics to transform it to Intellij annotation.
+            for (Diagnostic diagnostic : ds.getDiagnostics()) {
+                codeActionsLoading = codeActionsLoading | createAnnotation(diagnostic, document, ds, holder);
```
evie-lau, line 68: Not sure if you have a test scenario for this, but I'm using a language server that does not support codeactions, and is triggering a couple exceptions that are difficult to track down. 
But it seems like it starts from `createAnnotation` returning true here.

evie-lau: After modifying some logic here, it's still happening, so it may be that the `codeAction` is called from somewhere else without checking if the language server supports it?

angelozerr: Indeed I don't take care of that, plesase review my PR at https://github.com/redhat-developer/intellij-quarkus/pull/827

Issue #796, opened 2023-04-13
Codelens for REST resource is not shown if @GET and method are on the same line


adietish:
related to #795 

**Steps:**
1. ASSERT: have a quarkus project with a REST resource class where a method is annotated with a path.
ex.
```
@RestClient
@Path("/extension")
public class ExtensionsResource {

    @Inject
    ExtensionsService extensionsService;

    @GET @Path("/id/{id}") public Set<Extension> id(@PathParam String id) {
        return extensionsService.getById(id);
    }
}
```
1. EXEC: open the class in an editor

**Result:**
The url for the annotated method is NOT displayed. 
<img width="859" alt="image" src="https://user-images.githubusercontent.com/25126/231729293-36fe5084-2e40-4712-9976-3fe69af3c83c.png">

**Excpected result:**
The url should not be displayed like it is when the annotations and the method are on separate lines
<img width="859" alt="image" src="https://user-images.githubusercontent.com/25126/231729429-63855b06-f8f9-4f6a-a37e-e64cdf48b788.png">

angelozerr:
This issue should be fixed by lsp4ij

PR #824, opened 2023-04-27
feat: support codelens for Quarkus 3/Jakarta REST endpoints

### LINKED ISSUE #795: Codelens should not be indented
**Steps:**
1. ASSERT: have a quarkus project with a REST resource class where a method is annotated with a path.
ex.
```
@RestClient
@Path("/extension")
public class ExtensionsResource {

    @Inject
    ExtensionsService extensionsService;

    @GET
    @Path("/id/{id}")
    public Set<Extension> id(@PathParam String id) {
        return extensionsService.getById(id);
    }
}
```
1. EXEC: open the class in an editor

**Result:**
The url for the annotated method is displayed. The url is indented on the left though
image

**Excpected result:**
The url should not be indented, should ideally be left aligned with the `@Path` annotation

fbricon:
- feat: support codelens for Quarkus 3/Jakarta REST endpoints
- fix: align URL codelens with annotation (fixes #795)

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/jaxrs/JaxRsContext.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/jaxrs/JaxRsContext.java
@@ -13,6 +13,7 @@
 *******************************************************************************/
 package com.redhat.devtools.intellij.lsp4mp4ij.psi.core.jaxrs;
 
+import com.intellij.openapi.module.Module;
 import com.intellij.psi.PsiClass;
 import com.intellij.psi.search.searches.AnnotatedElementsSearch;
 import com.intellij.util.Query;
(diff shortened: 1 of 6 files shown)


angelozerr:
The PR works like a charm, the fix of codelens URL is better but there are some usecase which doesn't work like:

image

or:

image

But it is not a commons usecase. Commons usecases works great. We could improve again those CodeLens in a sperate PR.

I think we should see how to IJ manages `Usage inlay hint`.

fbricon:
> The PR works like a charm, the fix of codelens URL is better but there are some usecase which doesn't work like:
> 
> image
> 
> or:
> 
> image
> 
> But it is not a commons usecase. Commons usecases works great. We could improve again those CodeLens in a sperate PR.
> 
> I think we should see how to IJ manages `Usage inlay hint`.

This is already reported as #796. I'm working on a fix but it's a bit tricky.

angelozerr:
My bad CI build failed because tests https://github.com/redhat-developer/intellij-quarkus/actions/runs/4821889147/jobs/8588384324#step:5:79

@fbricon if you have time to fix it it would be really nice. Thanks.

fbricon:
@angelozerr sorry, the fix is here: https://github.com/redhat-developer/intellij-quarkus/pull/826

BTW, the commits of this PR were separated on purpose, they were not supposed to be squashed.

PR #827, opened 2023-04-27
fix: consume LSP request codeActions only if language server supports it

angelozerr:
Consume LSP request codeActions only if language server supports it

Today we use LSP4IJ with MicroProfile and Qute LS which have code actions capability. But if the language server has not the codeAction capability it consumes the codeActions request which throws an UnsuportedOperationException.

This PR check if the language server has the code action capability. 

See https://github.com/redhat-developer/intellij-quarkus/pull/819#discussion_r1179687022

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPIJUtils.java
@@ -319,15 +319,10 @@ public class LSPIJUtils {
         ApplicationManager.getApplication().runWriteAction(() -> edits.forEach(edit -> applyEdit(editor, edit, document)));
     }
 
-    public static boolean hasCapability(Either<Boolean, ? extends Object> eitherCapability) {
-        if(eitherCapability != null) {
-            if (eitherCapability.isLeft()) {
-                return eitherCapability.getLeft();
-            } else {
-                return eitherCapability.getRight() != null;
-            }
-        } else {
+    public static boolean hasCapability(final Either<Boolean, ? extends Object> eitherCapability) {
+        if(eitherCapability == null) {
             return false;
         }
+        return eitherCapability.isRight() || (eitherCapability.isLeft() && eitherCapability.getLeft());
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/operations/diagnostics/LSPDiagnosticsForServer.java
```diff
@@ -61,6 +63,7 @@ public class LSPDiagnosticsForServer {
 
     public LSPDiagnosticsForServer(LanguageServerWrapper languageServerWrapper, VirtualFile file) {
         this.languageServerWrapper = languageServerWrapper;
+        this.codeActionSupported = LSPIJUtils.hasCapability(languageServerWrapper.getServerCapabilities().getCodeActionProvider());
```
fbricon, line 66: languageServerWrapper.getServerCapabilities() != null && LSPIJUtils.hasCapability(languageServerWrapper.getServerCapabilities().getCodeActionProvider());

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPIJUtils.java
```diff
@@ -319,15 +319,10 @@ public static void applyEdits(Editor editor, Document document, List<TextEdit> e
...
-                return eitherCapability.getLeft();
-            } else {
-                return eitherCapability.getRight() != null;
-            }
-        } else {
+    public static boolean hasCapability(final Either<Boolean, ? extends Object> eitherCapability) {
+        if(eitherCapability == null) {
             return false;
         }
+        return eitherCapability.isRight() || eitherCapability.getLeft();
```
fbricon, line 326: if eitherCapability is neither right nor left, then returning eitherCapability.getLeft() will NPE on unboxing

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPIJUtils.java
```diff
@@ -319,15 +319,10 @@ public static void applyEdits(Editor editor, Document document, List<TextEdit> e
...
-                return eitherCapability.getLeft();
-            } else {
-                return eitherCapability.getRight() != null;
-            }
-        } else {
+    public static boolean hasCapability(final Either<Boolean, ? extends Object> eitherCapability) {
+        if(eitherCapability == null) {
             return false;
         }
+        return eitherCapability.isRight() || (eitherCapability.isLeft() && eitherCapability.getLeft());
```
fbricon, line 326: if you want to be thorough: 
> return (eitherCapability.isRight() && eitherCapability.getRight() != null  &&  eitherCapability.getRight()) || (eitherCapability.isLeft() && eitherCapability.getLeft() != null  &&  eitherCapability.getLeft());

angelozerr: I dont think it is correct since getObject is not à boolean. The main idea is to check if left is true or right is filled.

fbricon: OK, fair enough

PR #833, opened 2023-05-02
Dispose file wrapper when file is closed.

angelozerr:
Dispose file wrapper when file is closed.


adietish (approved):
fixes the issue for me. LGTM

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/ConnectDocumentToLanguageServerSetupParticipant.java
```diff
@@ -24,39 +25,45 @@
...
+public class ConnectDocumentToLanguageServerSetupParticipant implements ProjectManagerListener, FileEditorManagerListener {
 
-    public ConnectDocumentToLanguageServerSetupParticipant(Project project) {
-        this.project = project;
-    }
+    private static final Logger LOGGER = LoggerFactory.getLogger(ConnectDocumentToLanguageServerSetupParticipant.class);
 
     @Override
-    public void projectOpened() {
+    public void projectOpened(@NotNull Project project) {
```
evie-lau, line 36: Getting an error trying to test these changes with the runIde gradle task. 
```
ConnectDocumentToLanguageServerSetupParticipant.java:34: warning: [removal] projectOpened(Project) in ProjectManagerListener has been deprecated and marked for removal
    public void projectOpened(@NotNull Project project) {
                ^
```
Using IntelliJ 2023.1 with IJ Gradle plugin 1.13.3

angelozerr: I'm surprised to have this problem, because I'm using the same code than intellij-community https://github.com/search?q=repo%3AJetBrains%2Fintellij-community%20projectOpened&type=code

What is the proper mean to track projectOpened event?

angelozerr: Indeed in their Javadoc, there is https://github.com/JetBrains/intellij-community/blob/96e59f49bce9d19e20b4b9ad486123ce78cb4b0f/platform/projectModel-api/src/com/intellij/openapi/project/ProjectManagerListener.java#L20

After some researcj it seems ProjectActivity is the new API https://plugins.jetbrains.com/docs/intellij/plugin-components.html#project-open but it is not available for us.

angelozerr: As we cannot use this ProjectActivity APII, I reverse my PR to use the initial ProjectComponent (which is depreacted too)

adietish: @angelozerr: My general advice with API breaking in IDEA is to wait until the very last moment where the old API is NOT supported it any more. The above only is a warning of an upcoming removal.
The reasoning is the following: Ideally one should support latest version - 1y (better even more). According to jetbrains 80% of the users are within this range. Breaking API changes usually have a grace period where the old and the new API is supported. As soon as you switch to the new you need to bump your least supported IDEA version. Bridging APIs is usually possible (using reflection etc) but is ugly, little maintainable code that you want to avoid if possible.

angelozerr: Thanks @adietish for your feedback. As it worked, I keep the old API.

angelozerr:
CI build is working again! It was hard to understand why completion test failed but this test highlights the problem with restart language server problem. Nice to have this kind of test:)

PR #845, opened 2023-05-05
fix(skip-release): handle ideaVersion 'LATEST-EAP-SNAPSHOT'(#844)

### LINKED ISSUE #844: Latest IJ github action fails
```
* Where:
Build file '/home/runner/work/intellij-quarkus/intellij-quarkus/build.gradle' line: 73

* What went wrong:
A problem occurred evaluating root project 'intellij-quarkus'.
> For input string: "EST-"
```

adietish:
fixes #844

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,13 +70,15 @@ configurations {
 
 sourceSets {
     main  {
-        if (Integer.parseInt(ideaVersion.substring(3).substring(0, 4)) >= 2021) {
+        if (ideaVersion == "LATEST-EAP-SNAPSHOT"
+            || Integer.parseInt(ideaVersion.substring(3).substring(0, 4)) >= 2021) {
             java.srcDirs += 'src/main/java-post2021'
         } else {
             java.srcDirs += 'src/main/java-pre2021'
         }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -70,13 +70,15 @@ configurations {
 
 sourceSets {
     main  {
-        if (Integer.parseInt(ideaVersion.substring(3).substring(0, 4)) >= 2021) {
+        if (ideaVersion == "LATEST-EAP-SNAPSHOT"
```
angelozerr, line 73: I'm newbie with Gradle, but it looks like tit is Java code, should be not:
 
`"LATEST-EAP-SNAPSHOT".equals(ideaVersion)` ?

fbricon: This is `groovy`, i.e. less verbose Java ;-). `==` is similar to calling "equals", so this is fine.
https://stackoverflow.com/a/9682512/753170

adietish: it's groovy. In groovy `.equals()` is `==`. Reference equality in groovy is denoted with `===`
See https://docs.groovy-lang.org/latest/html/documentation/core-operators.html

PR #846, opened 2023-05-07
feat: Provide UI settings for Language Servers (#843)

### LINKED ISSUE #843: Provide UI settings for Language Servers
Provide UI settings for Language Servers to show list of language server definition. When a language server definition is selected it should provide the capability to configure:

 * debug port
 * suspend on debug
 * server trace level

angelozerr:
feat: Provide UI settings for Language Servers (#843)

Fixes #843


src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/settings/LanguageServerConfigurable.java
```diff
@@ -0,0 +1,105 @@
...
+    private LanguageServerView myView;
+
+    public LanguageServerConfigurable(LanguageServersRegistry.LanguageServerDefinition languageServerDefinition, Runnable updater) {
+        super(false, updater);
+        this.languageServerDefinition = languageServerDefinition;
+    }
+
+    @Override
+    public void setDisplayName(String name) {
+
```
adietish, line 45: maybe add a comment on why this is left empty on purpose. I see the super class calling it with the value of a text field when creating components.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/settings/LanguageServerConfigurable.java
```diff
@@ -0,0 +1,105 @@
...
+        settings.setDebugPort(myView.getDebugPort());
+        settings.setDebugSuspend(myView.isDebugSuspend());
+        LanguageServerSettingsState.getInstance().setLanguageServerSettings(languageServerDefinition.id, settings);
+    }
+
+    @Override
+    public void reset() {
+        LanguageServerSettingsState.LanguageServerDefinitionSettings settings = LanguageServerSettingsState.getInstance()
+                .getLanguageServerSettings(languageServerDefinition.id);
+        if (settings != null) {
```
adietish, line 95: shouldnt we at least log when no settings for a language server are found?

adietish (commented):
I finished the code review. Commented what I found.
I terms of testing it I have a few questions:
* I dont fully understand how it works: What happens If the servers are already started and I go and set the port on a sever that is already started? When trying to debug this I noticed repeated starting and stopping of `LanguageServerWrapper`. Is this intentional?
* In the UI afaik nothing prevents me from setting non-numerical values for the port.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/settings/LanguageServerConfigurable.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public boolean isModified() {
+        LanguageServerSettingsState.LanguageServerDefinitionSettings settings = LanguageServerSettingsState.getInstance()
+                .getLanguageServerSettings(languageServerDefinition.id);
+        if (settings == null) {
+            return true;
+        }
+        boolean modified = !myView.getDebugPort().equals(settings.getDebugPort());
+        modified = modified | myView.isDebugSuspend() != settings.isDebugSuspend();
```
fbricon, line 79: You don't need/want the right-hand side to be evaluated in this case. There is *practically* never a good reason to use single `|` over short-circuit `||`. 
See https://stackoverflow.com/a/7102013/753170

angelozerr:
> I dont fully understand how it works: What happens If the servers are already started and I go and set the port on a sever that is already started? 

Indeed it is not robust, we should stop the language server when debug port is filled, but as today starting of language server is not robust, I would like to do that in a separate PR. Could you create an issue for that.

> When trying to debug this I noticed repeated starting and stopping of LanguageServerWrapper. Is this intentional?

It is again a problem of start/stop of language server. I plan to fix this kind of problem inmy LSP console issue.

> In the UI afaik nothing prevents me from setting non-numerical values for the port.

Indeed but I don't want to spend some time in this kind of validation. I creates this UI settings page to support Trace which is more important than this kind of validation. Please create an issue for that.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LanguageServersRegistry.java
```diff
@@ -29,10 +29,12 @@ public abstract static class LanguageServerDefinition {
         public final @Nonnull String label;
         public final boolean isSingleton;
         public final @Nonnull Map<Language, String> languageIdMappings;
+        public String description;
```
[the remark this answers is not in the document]
angelozerr: fixed

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LanguageServersRegistry.java
```diff
@@ -224,6 +231,12 @@ private Set<LanguageServerDefinition> getAvailableLSFor(Language language) {
         return res;
     }
 
+    public Set<LanguageServerDefinition> getAllDefinitions() {
+        return connections
+                .stream()
+                .map( c -> (LanguageServerDefinition) c.getValue())
```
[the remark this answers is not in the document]
angelozerr: Good catch. fixed

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/server/JavaProcessCommandBuilder.java
```diff
@@ -0,0 +1,103 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
```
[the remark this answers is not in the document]
angelozerr: fixed

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/settings/LanguageServerConfigurable.java
angelozerr: fixed with comment.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/settings/LanguageServerConfigurable.java
angelozerr: Why you want to do that? I'm not sure it is relelvant to log that and it will polluate logs.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/settings/LanguageServerConfigurable.java
angelozerr: My bad, I wanted to mimic https://github.com/redhat-developer/intellij-tekton/blob/623343da0bd715cdfa0ab023066975e99b588cd4/src/main/java/com/redhat/devtools/intellij/tektoncd/settings/SettingsConfigurable.java#L38

Now I'm using ||

src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/settings/LanguageServerConfigurable.java
adietish: ok, so if this is not unexpected, then it's ok

adietish (approved):
LGTM

Issue #735, opened 2023-02-01
QuarkusTools cannot find a virtual file over a WSL2 environment


rmmarquini:
My dev env:
- Win10 + WSL2 + Ubuntu 22.04.1 LTS
- IntelliJ 2022.3.2 Ultimate Edition
- GraalVM 21.3.0 (build 11.0.13+7-jvmci-21.3-b05)

Every time I open an aplication.properties file IntelliJ shows me an error related to the QuarkusTools plugin, as described below:

Internal error: java.lang.UnsupportedOperationException: Cannot find virtual file for 'file:////wsl$/Ubuntu-22.04/home/myusr/Workspaces/w1/projects/monitor-summary-api/src/main/resources/application.properties'

java.util.concurrent.CompletionException: java.lang.UnsupportedOperationException: Cannot find virtual file for 'file:////wsl$/Ubuntu-22.04/home/myusr/Workspaces/w1/projects/monitor-summary-api/src/main/resources/application.properties'
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:708)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
	at com.redhat.devtools.intellij.quarkus.lsp.QuarkusLanguageClient.lambda$runAsBackground$2(QuarkusLanguageClient.java:124)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:186)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$executeProcessUnderProgress$13(CoreProgressManager.java:604)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:679)
	at com.intellij.openapi.progress.impl.CoreProgressManager.computeUnderProgress(CoreProgressManager.java:635)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:603)
[...]

PR #857, opened 2023-05-11
fix: remove usage of VfsUtil.findFileByURL

### LINKED ISSUE #856: Some codelens don't work in WSL2
The `Open template` code lens from a Java file, or the `Open type` header code lens in a qute template don't work, when the project is opened from a WSL2 environment. No exceptions. Just don't work.

The REST endpoint codelens work though.

Might be something related to #735, although sneakier.

fbricon:
The problem (and the fix) is similar to #735
Fixes #856

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPIJUtils.java
@@ -15,10 +15,7 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.roots.ProjectFileIndex;
 import com.intellij.openapi.util.TextRange;
-import com.intellij.openapi.vfs.LocalFileSystem;
-import com.intellij.openapi.vfs.VfsUtil;
-import com.intellij.openapi.vfs.VfsUtilCore;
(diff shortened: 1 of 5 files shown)


src/main/java/com/redhat/devtools/intellij/quarkus/lsp4ij/LSPIJUtils.java
```diff
@@ -248,11 +241,7 @@ public static VirtualFile findResourceFor(URI uri) {
     }
 
     public static VirtualFile findResourceFor(String uri) {
-        try {
-            return VfsUtil.findFileByURL(new URL(uri));
-        } catch (MalformedURLException e) {
-            return null;
-        }
+        return VirtualFileManager.getInstance().findFileByUrl(VfsUtilCore.fixURLforIDEA(uri));
```
angelozerr, line 244: The fix that you did [PsiUtilsLSImpl.java](https://github.com/redhat-developer/intellij-quarkus/commit/16c76e76e34a4a9535e0ef38289b88a42a5d4f16#diff-78d1994d7563bb33170f619c2358419c78997895ce3284cc2aaa0bfe8827ba79) should use `LSPIJUtils.findResourceFor` to have this fix in one location.

fbricon: done

PR #864, opened 2023-05-17
fix: Qute data model in template is not refreshed when Java file changed

### LINKED ISSUE #859: Qute data model in template is not refreshed when Java file changed
* Get the project https://github.com/adietish/code-with-quarkus
 * Open [SomePage.java](https://github.com/adietish/code-with-quarkus/blob/main/src/main/java/org/acme/SomePage.java) and [page.qute.html](https://github.com/adietish/code-with-quarkus/blob/main/src/main/resources/templates/page.qute.html)
 * add `{name}` in the HTML page:

image

 * You should see no error. Update the "name" in Java file with "nameX" and save the Java file. The Qute template should show an error but it doesn't work:

image

After the save we should have an error in Qute template with some quickfixes:

image

angelozerr:
fix: Qute data model in template is not refreshed when Java file changed (#859)

Fixes #859


angelozerr:
Here a demo with the fix:

IJQuteDataModelChanged

angelozerr:
Close for https://github.com/redhat-developer/intellij-quarkus/pull/875

PR #868, opened 2023-05-18
fix: Java file diagnostics is not refreshed when properties file is saved (#860)

### LINKED ISSUE #860: Java file diagnostics is not refreshed when properties file is saved
See https://github.com/OpenLiberty/liberty-tools-intellij/issues/328

IBM did a fix fir this issue https://github.com/OpenLiberty/liberty-tools-intellij/pull/385 but it breaks LSP support.

As we have LSP console soon it will be easy to compare LSP notifications request responses with vscode

angelozerr:
fix: Java file diagnostics is not refreshed when properties file is saved (#860)

Fixes #860


angelozerr:
Close for https://github.com/redhat-developer/intellij-quarkus/pull/875

PR #877, opened 2023-05-23
fix: Collect data model for Qute doesn't support fluent API (#876)

### LINKED ISSUE #876: Collect data model for Qute doesn't support fluent API
Given this Java class which defines several call expression data with fluent API:

```java

@Inject
private Template hello;

@GET
@Produces(MediaType.TEXT_HTML)
public TemplateInstance get(@QueryParam("name") String name) {
	return hello.data("height", 1.50, "weight", 50L)
			.data("age", 12)					
			.data("name", name);
}
```

The age and name are note collected in the data model.

angelozerr:
Collect data model for Qute doesn't support fluent API (#876)

Fixes #876


angelozerr:
> OK to merge once conventional commit issue is fixed

It should be fixed

PR #889, opened 2023-05-24
fix: StreamConnectionProvider instance should not be a singleton (#888)

### LINKED ISSUE #888: StreamConnectionProvider instance should be not a singleton
Today StreamConnectionProvider is a singleton per project. It causes the following problem which is hard to reproduce:

 * the language server is stopping (ex: you close all files)
 * you open a file and it restarts the language server. The old shutdown is not finished in async mode https://github.com/redhat-developer/intellij-quarkus/blob/dfec3374335593026b7fa478a56b03fc1ce0edba/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java#L435 and badly there is a TimeoutException (5sec).
 * the new language server is started.
 * After the TimeOut of 5sec, it stops the singleton StreamConnectionProvider https://github.com/redhat-developer/intellij-quarkus/blob/dfec3374335593026b7fa478a56b03fc1ce0edba/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java#L457 which stops the new process created by the new start of language server
 * As the new process is stopped, there are a lot of error with "Piped Stream is closing error." 
 * the launchFuture in this case is done which means that language server is NOT active and stop again the restarted language server https://github.com/redhat-developer/intellij-quarkus/blob/dfec3374335593026b7fa478a56b03fc1ce0edba/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java#L216

In otherwise,when shutdown takes 5sec, it provides the following problems:

 * the language server which is restarted, stops again
 * the old process of the StreamConnectionProvider is not killed.

To fix this issue the StreamConnectionProvider should be not a singleton.

To reproduce easily the problem, you need to emulate 5sec, you add https://github.com/redhat-developer/intellij-quarkus/blob/dfec3374335593026b7fa478a56b03fc1ce0edba/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java#L446 the following code:

 ```java
try {
                Thread.sleep(5000);
            } catch (InterruptedException e) {
                e.printStackTrace();
            }
```

You open a file, you close it and you reopen a file,the language server is started and stoppedagain.

### PR DESCRIPTION
**Pull request by angelozerr**
fix: StreamConnectionProvider instance should not be a singleton (#888)

Fixes #888

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServersRegistry.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServersRegistry.java
@@ -31,7 +31,6 @@ import java.util.stream.Collectors;
 
 /**
  * Language server registry.
- *
  */
 public class LanguageServersRegistry {
     private static final Logger LOGGER = LoggerFactory.getLogger(LanguageServersRegistry.class);
@@ -52,7 +51,7 @@ public class LanguageServersRegistry {
             this.label = label;
             this.description = description;
             this.isSingleton = isSingleton;
-            this.lastDocumentDisconnectedTimeout = lastDocumentDisconnectedTimeout != null  && lastDocumentDisconnectedTimeout > 0 ? lastDocumentDisconnectedTimeout : DEFAULT_LAST_DOCUMENTED_DISCONNECTED_TIMEOUT;
+            this.lastDocumentDisconnectedTimeout = lastDocumentDisconnectedTimeout != null && lastDocumentDisconnectedTimeout > 0 ? lastDocumentDisconnectedTimeout : DEFAULT_LAST_DOCUMENTED_DISCONNECTED_TIMEOUT;
             this.languageIdMappings = new ConcurrentHashMap<>();
         }
 
@@ -65,7 +64,7 @@ public class LanguageServersRegistry {
             return label != null ? label : id;
         }
 
-        public abstract StreamConnectionProvider createConnectionProvider();
+        public abstract StreamConnectionProvider createConnectionProvider(Project project);
 
         public LanguageClientImpl createLanguageClient(Project project) {
             return new LanguageClientImpl(project);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServersRegistry.java [hunk 1]
```diff
@@ -45,10 +44,13 @@ public abstract static class LanguageServerDefinition {
...
-        public final @Nonnull String id;
-        public final @Nonnull String label;
+        public final @Nonnull
+        String id;
+        public final @Nonnull
+        String label;
         public final boolean isSingleton;
-        public final @Nonnull Map<Language, String> languageIdMappings;
+        public final @Nonnull
+        Map<Language, String> languageIdMappings;
```
fbricon, line 53: can you keep the declarations on a single line?

src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServersRegistry.java
```diff
@@ -94,9 +96,15 @@ public ExtensionLanguageServerDefinition(ServerExtensionPointBean element) {
         }
 
         @Override
-        public StreamConnectionProvider createConnectionProvider() {
+        public StreamConnectionProvider createConnectionProvider(Project project) {
+            String serverImpl = extension.getImplementationClassName();
+            if (serverImpl == null || serverImpl.isEmpty()) {
+                throw new RuntimeException(
+                        "Exception occurred while creating an instance of the stream connection provider, you have to define @class attribute"); //$NON-NLS-1$
```
fbricon, line 103: "you have to define @class attribute" ??? where?

src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServersRegistry.java
```diff
@@ -222,9 +232,12 @@ public List<ContentTypeToLanguageServerDefinition> getContentTypeToLSPExtensions
...
 
-        @Nonnull public final String id;
-        @Nonnull public final Language language;
-        @Nullable public final String languageId;
+        @Nonnull
+        public final String id;
+        @Nonnull
+        public final Language language;
+        @Nullable
+        public final String languageId;
```
fbricon, line 240: keep 1-line declarations

fbricon (changes requested):

src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServersRegistry.java
angelozerr: fixed, but it is format code of IJ which does that

PR #894, opened 2023-05-26
fix: LSP console shouldn't be hidden when IJ is indexing

fbricon:
Keep LSP console displayed when IJ is indexing, by making LSPConsoleToolWindowFactory implement DumbAware
See https://plugins.jetbrains.com/docs/intellij/tool-windows.html#contents-tabs


angelozerr:
Thanks @fbricon !

PR #897, opened 2023-05-30
fix: Fragment support between Java / Template file (#896)

### LINKED ISSUE #896: Fragment support between Java / Template file
Open https://github.com/ia3andy/renotes/blob/98b21e4b465d49171f22e11296f6a1bc6ff0e75b/src/main/java/rest/Notes.java#L25

You should see:

image

But today there is an error validation on `notes$noteList` method because as fragment is not supported it tries to open the `notes$noteList.html` file instead of `notes.html`

See commit in Quarkus LS https://github.com/redhat-developer/quarkus-ls/commit/820e5a9b71302e9f21ea5bc29e1610b51b4757d7

angelozerr:
fix: Fragment support between Java / Template file (#896)

Fixes #896

Port of commit https://github.com/redhat-developer/quarkus-ls/commit/820e5a9b71302e9f21ea5bc29e1610b51b4757d7 of Quarkus LS

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -161,8 +161,8 @@ dependencies {
     lsp ('com.redhat.microprofile:com.redhat.quarkus.ls:0.13.0') {
         transitive = false
     }
-    implementation 'com.redhat.microprofile:com.redhat.qute.ls:0.13.0'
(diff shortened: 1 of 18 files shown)


angelozerr:
@fbricon is this PR is working for you? The CI build fais only for macos

fbricon (approved):

Issue #823, opened 2023-04-27
NPE in lsp4mpij when fieldType in MicroProfileRestClientDiagnosticsParticipant.validateField() is null


adietish:
Found the following when testing #819:

When debugging why an "Insert @RestClient" wasn't suggested we found an NPE at the following line:

[MicroProfileRestClientDiagnosticsParticipant.validateField()](https://github.com/redhat-developer/intellij-quarkus/blob/main/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/restclient/java/MicroProfileRestClientDiagnosticsParticipant.java#L127):
```java
	private static void validateField(PsiField field, List<Diagnostic> diagnostics, JavaDiagnosticsContext context) {
		String uri = context.getUri();
		DocumentFormat documentFormat = context.getDocumentFormat();
		boolean hasInjectAnnotation = AnnotationUtils.hasAnyAnnotation(field, INJECT_JAVAX_ANNOTATION, INJECT_JAKARTA_ANNOTATION);
		boolean hasRestClientAnnotation = AnnotationUtils.hasAnnotation(field, REST_CLIENT_ANNOTATION);
		String fieldTypeName = PsiTypeUtils.getResolvedTypeName(field);
		PsiClass fieldType = PsiTypeUtils.findType(field.getManager(), fieldTypeName);
		boolean hasRegisterRestClient = AnnotationUtils.hasAnnotation(fieldType, REGISTER_REST_CLIENT_ANNOTATION)
--->			&& fieldType.isInterface();
```
It is unclear why `fieldType` would be `null` after a new start of the IDE. We suspect the IJ indexing.

**Steps:**
1. ASSERT: have IDEA freshly started
1. ASSERT: have https://github.com/adietish/code-with-quarkus present as project in IDEA
2. EXEC: open `ExtensionsResource` in an editor
3. ASSERT: make sure indexing is finished
4. EXEC: go to `extensionsService` instance variable and hit Alt + Enter
5. EXEC: inspect quickfixes to see if "Insert @RestClient" is suggested

**Result:**
It is not suggested even thought the inlay hint for the `id` method is present.
![image](https://user-images.githubusercontent.com/25126/234898114-5bab526f-e17c-44a8-87d6-521d53d280cf.png)

**Expected Result:**
![image](https://user-images.githubusercontent.com/25126/234898318-9106502b-55e4-45e6-ae3c-9839590516e4.png)

[...]

PR #902, opened 2023-05-31
fix: prevent potential NPE in MicroProfileRestClientDiagnosticsParticipant.validateField()

fbricon:
No idea how the NPE could happen, but this fix guards against it.
Fixes https://github.com/redhat-developer/intellij-quarkus/issues/823


angelozerr:
LGTM

PR #903, opened 2023-05-31
fix: Fix Qute test

angelozerr:
fix: Fix Qute test


angelozerr:
@fbricon I created an issue at https://github.com/redhat-developer/quarkus-ls/issues/875 about this usecase.

PR #906, opened 2023-06-01
fix: Perform updates to Stop Language Server Timer synchronously. (#905)

### LINKED ISSUE #905: Perform updates to Stop Language Server Timer synchronously.
See https://github.com/eclipse/lsp4e/issues/644

angelozerr:
fix: Perform updates to Stop Language Server Timer synchronously. (#905)

Fixes #905


angelozerr:
Please don't merge, I have seen some trouble when we start stop ls, close files. I will investigate the problem when I will have time.

More there is a problem with thread lock in test https://github.com/redhat-developer/intellij-quarkus/actions/runs/5144122828/jobs/9260013480?pr=906#step:5:463

I think this PR is not very important for now.

PR #943, opened 2023-06-12
build: bump dev IDEA version to 2022.1

fbricon:
This will help catch problems that only exists in newer versions of IDEA


fbricon:
PR needs to adapt ./pulltests.sh to the new version, so the test infra can compile

PR #976, opened 2023-06-26
fix: Start/stop actions from language server explorer do nothing.

angelozerr:
When you click on Start/stop actions from language server explorer, it does nothing.

image


angelozerr:
Tested on IC-2021.3 and EAP

PR #986, opened 2023-06-30
feat: update to lsp4mp 0.8.0  and quarkus-ls 0.15.0

### LINKED ISSUE #962: Update quarkus-ls to 0.15.0 and lsp4mp to 0.8.0
quarkus-ls 0.15.0 and lsp4mp 0.8.0 were released, in order to update, we need to **at least** make QuarkusLanguageClient implement the now non-default methods, as per:

- https://github.com/eclipse/lsp4mp/issues/398
- https://github.com/redhat-developer/quarkus-ls/issues/870

fbricon:
Fixes #962


src/main/java/com/redhat/devtools/intellij/quarkus/lsp/QuarkusLanguageClient.java
```diff
@@ -185,4 +190,21 @@ public CompletableFuture<CodeAction> resolveCodeAction(CodeAction unresolved) {
       return (CodeAction) PropertiesManagerForJava.getInstance().resolveCodeAction(unresolved, PsiUtilsLSImpl.getInstance(getProject()));
     });
   }
+
+  @Override
+  public CompletableFuture<JavaCursorContextResult> getJavaCursorContext(MicroProfileJavaCompletionParams params) {
```
angelozerr, line 195: I think this method is not used in MP LS, let's keep like this, but I think we should remove it,see https://github.com/eclipse/lsp4mp/issues/408

angelozerr:
Thanks so much @fbricon !

PR #1005, opened 2023-07-11
fix: Ensure config_ordinal only appears in MicroProfile projects

fbricon:
Port of https://github.com/eclipse/lsp4mp/commit/caf91c68e7f2fe27072f1f947e3c5476f3289d8f

Part of #806

To test, open a non-quarkus project (eg. Lemminx), add an `src/main/resources/application.properties` file and trigger completion on an empty line. config_ordinal should not appear.


angelozerr:
Thanks @fbricon !

PR #1011, opened 2023-07-12
fix: remove extra colon in inlay hints

fbricon:
Qute Inlay hints are displayed with a double "::"

Screenshot 2023-07-12 at 15 18 01

The non-command part of the inlay hint was added twice. This PR fixes it:

Screenshot 2023-07-12 at 16 38 08


angelozerr:
Thanks @fbricon !

Issue #1010, opened 2023-07-12
Wrong location of the caret after apply completion


angelozerr:
See the following demo:

![CaretProblem](https://github.com/redhat-developer/intellij-quarkus/assets/1932211/666cbeae-0657-434d-8c90-b41a5f339137)

The caret is working well if you apply completion without typing something. If you open completion and you type something and you apply completion, caret is not set correctly which breaks the completion flow.

angelozerr:
Fixed with https://github.com/redhat-developer/intellij-quarkus/pull/1018

PR #1018, opened 2023-07-15
fix: Wrong location of the caret after apply completion #1010

angelozerr:
fix: Wrong location of the caret after apply completion #1010

Fixes (#1010)


angelozerr:
This PR fixes the https://github.com/redhat-developer/intellij-quarkus/issues/1010 issue and fix too the snippet completion which now replaces the content declared in the TextEdit, see the following demo which remove the original content `{#}` after snippet completion is applied (in the master it keeps `{#}` and after apply completion you have `{#{#for .....}}`, now it generates  `{#for .....}`:

FixSnippetConmpletiontextEdit

PR #1025, opened 2023-07-16
fix: Freeze on parameter declaration completion (#1015)

### LINKED ISSUE #1015: Freeze on parameter declaration completion
Given this Qute template:

`{@|}`

Open the completion at `|` completion freezes IJ.

angelozerr:
fix: Freeze on parameter declaration completion (#1015)

Fixes #1015

It should be fast if you declare the package (ex : `{@java.lang.|`) and little slow if you type directly the class without package (ex : `{@Str|}`. See the following demo:

ParameterDeclarationInIJ

Perhaps it is slow because we return too many items? But it is now useable.


fbricon:
"java.lang." prefix should be omitted from the resulting results

angelozerr:
> "java.lang." prefix should be omitted from the resulting results

Why? I know that parameter declaration must be improved, but main goal was to improve performance to use it. Please create new issues with your idea to improve this support.

PR #1040, opened 2023-07-24
fix: quarkus.kubernetes.* properties not shown or validated (#858)

### LINKED ISSUE #858: quarkus.kubernetes.* properties not shown or validated
However `quarkus.kubernetes-client.*` properties are shown in the autocomplete

angelozerr:
fix: quarkus.kubernetes.* properties not shown or validated (#858)

Fixes #858

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/classpath/ClasspathResourceChangedListener.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/classpath/ClasspathResourceChangedListener.java
@@ -56,70 +56,8 @@ class ClasspathResourceChangedListener extends PsiTreeChangeAdapter implements B
 
     private final ClasspathResourceChangedManager manager;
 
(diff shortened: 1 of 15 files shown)


angelozerr:
This PR fixes the issue. I had to do a big refactoring because before the compute of Quarkus deployment for a given module was done each time a library was added/removed. It means that when project was loaded the first time, the index process was started and on each library (maven dependecy) it started the process of compute of Quarkus deployment. You could have 100 compute of Quarkus deploymenet if your maven project had 100 dependencies.

As it is not possible to detect a start/stop of project loading (it is possible with WorkspaceModelTopics but this API changed of package and API on several versions of IJ, so it is not possible to use it), we need to track again when library is added/removed, but now we manage debounce of this change, we avoid to load Quarkus deployment on each library update,just on last library update).

angelozerr:
Here a demo:

KubernetesPropertiesSupportInIJ

PR #1051, opened 2023-07-26
feat: Support crash of language server starting (#1001)

### LINKED ISSUE #1001: Support crash of language server starting
When a language server cannot be stated it freezes the UI. We should avoid freezing the UI and retry the start of the language server 5 times for instance.

After 5 times, the labguage server become disabled. To re-enable it we can use the existing Restart action from the contextual menu of the console.

We should show those states with icon error and error message when language server cannot be started.

I think this issue is important because when language server cannot be started, it freezes IJ every time.

angelozerr:
feat: Support crash of language server starting (#1001)

Fixes #1001


angelozerr:
Here a demo with:

 * Qute LS which can starts correctly(the path of Qute LS JAR is bad)
 * MicroProfile LS which starts correctly

The demo show in action the call of the Qute LS which fails, the LS is called for many LSP requests. I set the max amounts to 10.
You cansee in the demo that it tries to start the Qute LS 10 times and after that it disable the Qute LS. To reenable it you can do a Restart.

LanguageServerCracsh

I add too a new action `Copy Start Command` to copy in the clipborad the java command which start the process. It was easy to implement it and I was tired to debug this command each time.

I have tested the PR with different usecase:

 * the start of LS cannot be done
 * the start can be done but the process is killed after 

and we have no IJ freeze.

angelozerr:
This PR is working correctly but perhaps we will nned to improve again the exception because sometimes (when server crashes) I see a CannotStartServerException in the log but it is hard to reproduce it.

To test this PR:

 * emulate bad command by updating with bad path the Qute uber jar https://github.com/redhat-developer/intellij-quarkus/blob/357eee52a7668ce6a1970e1a89278c37201e3b67/src/main/java/com/redhat/devtools/intellij/qute/lsp/QuteServer.java#L34
 * start the Qute LS normally (open a Qute template file). Set a breakpoint here https://github.com/redhat-developer/intellij-quarkus/blob/dd7adb607bd181f4ac7d7410c0c4443ef5dae073/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java#L304 (when LSP request should be consumed by the process). Open a completion it wait for your breakpoint.
Kill the process (use pid that you can see in the LSP console) and continue the run. You should have a Pipe exception but the server should be stopped and restarted again with a new pid without freezing IJ.

PR #1064, opened 2023-07-31
build: raise minimal compatible IJ IDEA version to 2022.2

fbricon:


angelozerr:
Thanks @fbricon !

PR #1067, opened 2023-08-01
fix: go-to-definition and inlay hint links for JDK classes

### LINKED ISSUE #1063: Open type from inlayhint doesn't work in Qute template
Clicking on the inlay hint does nothing, despite the displayed hint to open the type.
Screenshot 2023-07-31 at 17 01 02

Tested on https://github.com/ia3andy/quarkus-blast/blob/main/src/main/resources/templates/GameController/partials/grid.html#L4

fbricon:
- JDK classes use a different URI protocol (`jrt:`)
- Don't use PsiNameIdentifierOwner.getNameIdentifier().getNavigationElement() as it opens the .class file instead of the source, if available

Fixes #1063


angelozerr:
It works like a charm, great job @fbricon !

PR #1070, opened 2023-08-02
fix: Closing one module window doesn't shutdown the running LS instances

### LINKED ISSUE #1069: Closing one module window doesn't shutdown the running LS instances
- Open a Java project
- in a terminal, run `jps -mlv`, you should see 2 Application processes running (qute-ls and lsp4mp)
- close the module window
- Open another Java project
- in a terminal, run `jps -mlv`, you should now see 4 Application processes running (2 qute-ls and 2 lsp4mp)
- No matter how long you wait the 4 processes are still alive while IJ is opened (parent process watcher doesn't kill the first 2 because the parent process is still alive)

angelozerr:
fix: Closing one module window doesn't shutdown the running LS instances

Fixes #1069


fbricon:
Thanks @angelozerr!

PR #1076, opened 2023-08-03
perf: Start language server only when indexation is finished

### LINKED ISSUE #1075: Start language server only when indexation is finished
IJ indexation takes some times and language server are started even if IJ indexation is occured. This start is not very interesting because all features like completion, validation, etc are not available during indexation:

 * completion: even if completion is mark as DumbAware, it needs to consume IJ Java services which are not available during the indexation
 * validation: the IJ annotator validation seems never ready when IJ indexation occurs.

Starting language server after the IJ indexation should improve performance.

angelozerr:
perf: Start language server only when indexation is finished

Fixes #1075


src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServiceAccessor.java
```diff
@@ -42,6 +42,8 @@
  */
 public class LanguageServiceAccessor {
     private static final Logger LOGGER = LoggerFactory.getLogger(LanguageServiceAccessor.class);
+
+    public static final Predicate<ServerCapabilities> ANY_SERVER_CAPABILITIES = t-> true;
```
fbricon, line 46: unused?

angelozerr: fixed

Issue #794, opened 2023-04-13
Provide a preference page


fbricon:
Those  settings should be exposed in a preference page:

### Quarkus

> * `quarkus.tools.debug.terminateProcessOnExit` : Determines whether to terminate the quarkus:dev task after closing the debug session. Default is `Ask`.
> * `quarkus.tools.alwaysShowWelcomePage` : Determines whether to show the welcome page on extension startup. Default is `true`.
> * `quarkus.tools.starter.api` : Quarkus API base URL. Default is `https://code.quarkus.io/api`.
> * `quarkus.tools.starter.showExtensionDescriptions`: Determines whether to show the Quarkus extension descriptions when selecting Quarkus extensions. Default is `true`.
> * `quarkus.tools.propertiesLanguageMismatch` : Action performed when detected Quarkus properties have an incorrect language. Default is `forceQuarkus`.

### Microprofile

> * `microprofile.tools.formatting.surroundEqualsWithSpaces` : Insert spaces around the equals sign when formatting the application.properties file. Default is `false`.
> * `microprofile.tools.trace.server` : Trace the communication between VS Code and the MicroProfile Language Server in the Output view.
> * `microprofile.tools.symbols.showAsTree` : Show MicroProfile properties as tree (Outline). Default is `true`.
> * `microprofile.tools.validation.enabled` : Enables MicroProfile validation. Default is `true`.
> * `microprofile.tools.validation.duplicate.severity` : Validation severity for duplicate properties for MicroProfile `*.properties` files.
> Default is `warning`.
> * `microprofile.tools.validation.syntax.severity` : Validation severity for property syntax checking for MicroProfile `*.properties` files.
> Default is `error`.
> * `microprofile.tools.validation.required.severity` : Validation severity for required properties for MicroProfile `*.properties` files.
> Default is `none`.
> * `microprofile.tools.validation.expression.severity` : Validation severity for property expressions for MicroProfile `*.properties` files.
> Default is `error`.
> * `microprofile.tools.validation.unknown.severity` : Validation severity for unknown properties for MicroProfile `*.properties` files. Default is `warning`.
> * `microprofile.tools.validation.unknown.excluded` : Array of properties to ignore for unknown properties validation. Patterns can be used ('\*' = any string, '?' = any character).
> Default is `["*/mp-rest/providers/*/priority", "mp.openapi.schema.*", "kafka-streams.*", "camel.*"]`.
[...]

PR #1117, opened 2023-08-22
feat: expose MicroProfile server settings

fbricon:
Part of #794 
Screenshot 2023-08-22 at 14 53 36


angelozerr:
It is out of thescope of this PR, but I wonder if we should remove LSP ?

image

angelozerr:
It works like a charm.

IMHO I think we should improve description of each error like HTML does:

image

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/settings/ProblemSeverity.java
```diff
@@ -0,0 +1,46 @@
...
+package com.redhat.devtools.intellij.lsp4mp4ij.settings;
+
+import com.intellij.lang.annotation.HighlightSeverity;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Problem severity levels used by LSP4MP
+ */
+//TODO move to lsp4ij?
+public enum ProblemSeverity {
```
angelozerr, line 23: Why not using DiagnosticSeverity and null for none?

src/main/java/com/redhat/devtools/intellij/quarkus/lsp/QuarkusLanguageClient.java
```diff
@@ -74,6 +87,24 @@ private void sendPropertiesChangeEvent(List<MicroProfilePropertiesScope> scope,
         }
     }
 
+    @Override
+    public void profileChanged(@NotNull InspectionProfile profile) {
+        MicroProfileInspectionsInfo newInspectionState = MicroProfileInspectionsInfo.getMicroProfileInspectionInfo(getProject());
```
angelozerr, line 92: Explain that you are tracking MP validtaion settings (declared in Editor/Inspection/MicroProfue UI settings) to push config to the LS with didChangeConfig.

fbricon:
I also need to fix:
```
com.redhat.devtools.intellij.quarkus.completion.MavenApplicationPropertiesCompletionTest

  Test testBooleanCompletion FAILED (43.3s)

  junit.framework.AssertionFailedError: expected:<2> but was:<0>
      at com.redhat.devtools.intellij.quarkus.completion.MavenApplicationPropertiesCompletionTest.testBooleanCompletion(MavenApplicationPropertiesCompletionTest.java:41)
```

fbricon:
> It is out of thescope of this PR, but I wonder if we should remove LSP ?

Done, it provided no value.

src/main/java/com/redhat/devtools/intellij/quarkus/lsp/QuarkusLanguageClient.java
fbricon: done

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/settings/ProblemSeverity.java
fbricon: DiagnosticSeverity has other values like hint and info, which could be mapped to IJ's severities, but since we don't use them, eventually I kept ProblemSeverity, as it makes things simpler, less risks of confusion.

fbricon:
I removed the unassigned inspection, as LSP4MP ignores the unassigned severity (you can't set it neither in vscode)

I'm pondering whether to remove the duplicates inspection as well, since IJ sets the severity to error on its own, even though Properties Duplicates inspection is disabled (probably a bug in IJ)

angelozerr:
> WDYT?

It is great! The dark mode should be improved but if you wish we can do that in an another PR.

I think that we should use editor background (it means that it html should be dynamic, why not use Qute to load your HTML file and replace bagkground color with the IJ background color?)

fbricon:
This simpler approach works in both cases, I think:
Screenshot 2023-08-25 at 15 53 05
Screenshot 2023-08-25 at 15 53 34

angelozerr:
Indeed it is better, the yellow div border should use perhaps a gray color, because in dark mode I find it is a little flashy.

angelozerr:
@fbricon do you think it is possible and easy to do to add a link the the M`icroProfile / Properties` page to the `Inspections MicroProfile validation settings`?

fbricon:
> @fbricon do you think it is possible and easy to do to add a link the the M`icroProfile / Properties` page to the `Inspections MicroProfile validation settings`?

I actually thought about it then forgot to do it ;-)

fbricon:
Darker yellow:
Screenshot 2023-08-25 at 16 53 35
Screenshot 2023-08-25 at 16 53 55

angelozerr:
Thanks so much @fbricon !

Issue #1089, opened 2023-08-09
Too many non-blocking read actions submitted at once in LSPDiagnosticHandler


mfortunat:
Guys, the newly released version 1.24.0.508 has another, repeating, issue.
IJ version is: "Build #IU-232.8660.185, built on July 26, 2023"

Stack trace:
-----
java.lang.Throwable: Too many non-blocking read actions submitted at once. Please use coalesceBy, BoundedTaskExecutor or another way of limiting the number of concurrently running threads.: 11 with similar stack traces are currently active
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:370)
	at com.intellij.openapi.application.impl.SubmissionTracker.preventTooManySubmissions(SubmissionTracker.java:45)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$Submission.<init>(NonBlockingReadActionImpl.java:272)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl.submit(NonBlockingReadActionImpl.java:227)
	at com.redhat.devtools.intellij.lsp4ij.operations.diagnostics.LSPDiagnosticHandler.accept(LSPDiagnosticHandler.java:52)
	at com.redhat.devtools.intellij.lsp4ij.operations.diagnostics.LSPDiagnosticHandler.accept(LSPDiagnosticHandler.java:38)
	at com.redhat.devtools.intellij.lsp4ij.client.LanguageClientImpl.publishDiagnostics(LanguageClientImpl.java:68)
	at jdk.internal.reflect.GeneratedMethodAccessor291.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:65)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.notify(GenericEndpoint.java:152)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleNotification(RemoteEndpoint.java:220)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:187)
[...]

PR #1129, opened 2023-08-28
fix: Bind language servers to Project instead of Module

### LINKED ISSUE #891: Bind language servers to Project instead of Module

angelozerr:
fix: Bind language servers to Project instead of Module

Fixes #891

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java
@@ -14,7 +14,6 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.intellij.AppTopics;
-import com.intellij.ProjectTopics;
 import com.intellij.lang.Language;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.application.ApplicationInfo;
@@ -23,12 +22,10 @@ import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.EditorFactory;
 import com.intellij.openapi.editor.event.DocumentEvent;
 import com.intellij.openapi.editor.event.DocumentListener;
-import com.intellij.openapi.fileEditor.FileDocumentManager;
 import com.intellij.openapi.fileEditor.FileDocumentManagerListener;
 import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.fileEditor.FileEditorManagerListener;
 import com.intellij.openapi.module.Module;
-import com.intellij.openapi.project.ModuleListener;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Disposer;
 import com.intellij.openapi.vfs.VirtualFile;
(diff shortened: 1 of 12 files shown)


src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java
```diff
@@ -793,7 +793,7 @@ private CompletableFuture<LanguageServer> connect(@Nonnull URI absolutePath, Doc
 
         VirtualFile file = FileDocumentManager.getInstance().getFile(document);
         if (file != null && file.exists()) {
-            watchProject(LSPIJUtils.getProject(file), false);
+            watchProject(LSPIJUtils.getModule(file), false);
```
fbricon, line 796: maybe rename to watchModule

src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java
```diff
@@ -366,7 +366,7 @@ public synchronized void start() throws LanguageServerException {
                         final Map<URI, Document> toReconnect = filesToReconnect;
                         initializeFuture.thenRunAsync(() -> {
                             if (this.initialProject != null) {
-                                watchProject(this.initialProject, true);
+                                //watchProject(this.initialProject, true);
```
fbricon, line 369: why?

fbricon:
I tested it on https://github.com/quarkiverse/quarkus-ironjacamar, which is multimodule, didn't see any problem. LGTM

src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java
angelozerr: I remove this code since we don't support workspace folder.

src/main/java/com/redhat/devtools/intellij/lsp4ij/LanguageServerWrapper.java
angelozerr: I remove this code since we don't support workspace folder.

angelozerr:
@fbricon is ready now, I have removed all workspace folder support.

@mfortunat this PR should fix your problem with your multi-module https://github.com/redhat-developer/intellij-quarkus/issues/1089#issuecomment-1710121155

fbricon:
@angelozerr it looks like properties leak to unrelated projects. e.g. using https://github.com/quarkiverse/quarkus-ironjacamar

- set a new property in the integration test [application.properties]( https://github.com/quarkiverse/quarkus-ironjacamar/blob/main/integration-tests/artemis-jms/src/main/resources/application.properties)
- inject a ConfigProperty with the same name in [IronJacamarProcessor](https://github.com/quarkiverse/quarkus-ironjacamar/blob/main/deployment/src/main/java/io/quarkiverse/ironjacamar/deployment/IronJacamarProcessor.java#L61)

The downstream property is resolved on hover in the deployment module, even though it's not on that module's classpath

Screenshot 2023-09-07 at 18 55 50

angelozerr:
@fbricon I think this problem is the same with just one module. We have problem with cache https://github.com/redhat-developer/intellij-quarkus/issues/860

To test this PR,I suggest that you create an application.properties in each module and on each module you define a class with ConfigProperty and you should see the proper properties in completion,validation for each module

angelozerr:
> The downstream property is resolved on hover in the deployment module, even though it's not on that module's classpath

I think it is a feature coming from IJ himself, please check that with LSP console traces

fbricon:
thanks @angelozerr

PR #1135, opened 2023-08-31
feat: Mark observer methods as implicitly used

### LINKED ISSUE #1134: Mark observer methods as implicitly used
Seen on https://github.com/quarkusio/quarkus-quickstarts/blob/main/reactive-routes-quickstart/src/main/java/org/acme/reactive/routes/MyRouteRegistar.java#L11 :

Screenshot 2023-08-31 at 16 58 11

> An observer method is a non-abstract method of a managed bean class or session bean class (or of an extension). An observer method may be either static or non-static. If the bean is a session bean, the observer method must be either a business method of the EJB or a static method of the bean class.
Each observer method must have exactly one event parameter, of the same type as the event type it observes. Event qualifiers may be declared by annotating the event parameter. When searching for observer methods for an event, the container considers the type and qualifiers of the event parameter.
If the event parameter does not explicitly declare any qualifier, the observer method observes events with no qualifier.
The event parameter type may contain a type variable or wildcard.
In addition to the event parameter, observer methods may declare additional parameters, which may declare qualifiers. These additional parameters are injection points.
>
>  ```java
> public void afterLogin(@Observes LoggedInEvent event, @Manager User user, Logger log) { ... }
> ``` 
>A bean (or extension) may declare multiple observer methods.
Observer methods are inherited by bean subclasses.
Interceptors and decorators may not declare observer methods.

fbricon:
fixes #1134 

See https://github.com/quarkusio/quarkus-quickstarts/blob/main/reactive-routes-quickstart/src/main/java/org/acme/reactive/routes/MyRouteRegistar.java#L11:

Screenshot 2023-08-31 at 17 31 32

@gastaldi FYI

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/JavaEEImplicitUsageProvider.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/JavaEEImplicitUsageProvider.java
@@ -14,17 +14,24 @@
 package com.redhat.devtools.intellij.lsp4mp4ij.psi.core;
 
 import com.intellij.codeInsight.daemon.ImplicitUsageProvider;
-import com.intellij.psi.PsiElement;
+import com.intellij.psi.*;
 import com.redhat.devtools.intellij.lsp4mp4ij.psi.core.jaxrs.JaxRsConstants;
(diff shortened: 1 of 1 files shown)


gastaldi (approved):
Nice!!

angelozerr:
Thanks @fbricon !

PR #1151, opened 2023-09-08
updateQuteConfigCommand

fbricon:
- feat: add command to update microprofile configuration from LSP4MP
- feat: add command to update Qute configuration from Qute LS


fbricon:
something's wrong with the commits

PR #1154, opened 2023-09-11
feat: GraphQL improvements

jmartisk:
- Remove the check that forbids returning `void` from GraphQL operations. This is legal in recent Quarkus versions.
- Add some checks for the correct placement of GraphQL directives.

### DIFF
--- a/projects/lsp4mp/projects/maven/microprofile-graphql/pom.xml
+++ b/projects/lsp4mp/projects/maven/microprofile-graphql/pom.xml
@@ -40,6 +40,11 @@
         <version>1.0.2</version>
     </dependency>
     <dependency>
+      <groupId>io.smallrye</groupId>
+      <artifactId>smallrye-graphql-api</artifactId>
+      <version>2.4.0</version>
+    </dependency>
+    <dependency>
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-api</artifactId>
         <version>1.7.30</version>
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/TypeSystemDirectiveLocation.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/TypeSystemDirectiveLocation.java
@@ -11,21 +11,23 @@
 * Contributors:
 *     Red Hat Inc. - initial API and implementation
 *******************************************************************************/
-package com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.graphql.java;
-
-import com.redhat.devtools.intellij.lsp4mp4ij.psi.core.java.diagnostics.IJavaErrorCode;
+package com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.graphql;
 
 /**
- * MicroProfile GraphQL diagnostics error code.
+ * GraphQL schema element types - used for declaring the allowed
+ * placement of directives in a GraphQL schema.
+ * See http://spec.graphql.org/draft/#TypeSystemDirectiveLocation
  */
-public enum MicroProfileGraphQLErrorCode implements IJavaErrorCode {
-
-	WRONG_DIRECTIVE_PLACEMENT
-	;
-
-	@Override
-	public String getCode() {
-		return name();
-	}
-
+public enum TypeSystemDirectiveLocation {
+    SCHEMA,
+    SCALAR,
+    OBJECT,
+    FIELD_DEFINITION,
+    ARGUMENT_DEFINITION,
+    INTERFACE,
+    UNION,
+    ENUM,
+    ENUM_VALUE,
+    INPUT_OBJECT,
+    INPUT_FIELD_DEFINITION
 }
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
@@ -14,44 +14,176 @@
 package com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.graphql.java;
 
 
+import java.text.MessageFormat;
+import java.util.Arrays;
 import java.util.logging.Logger;
 
+import com.intellij.lang.jvm.JvmMethod;
+import com.intellij.lang.jvm.types.JvmArrayType;
+import com.intellij.lang.jvm.types.JvmPrimitiveType;
+import com.intellij.lang.jvm.types.JvmReferenceType;
+import com.intellij.lang.jvm.types.JvmType;
+import com.intellij.lang.jvm.types.JvmTypeVisitor;
+import com.intellij.lang.jvm.types.JvmWildcardType;
 import com.intellij.openapi.module.Module;
 import com.intellij.psi.PsiAnnotation;
+import com.intellij.psi.PsiAnnotationMemberValue;
+import com.intellij.psi.PsiArrayInitializerMemberValue;
+import com.intellij.psi.PsiArrayType;
+import com.intellij.psi.PsiClass;
+import com.intellij.psi.PsiClassType;
+import com.intellij.psi.PsiEnumConstant;
+import com.intellij.psi.PsiField;
+import com.intellij.psi.PsiJvmModifiersOwner;
 import com.intellij.psi.PsiMethod;
+import com.intellij.psi.PsiParameter;
+import com.intellij.psi.PsiParameterList;
 import com.intellij.psi.PsiType;
+import com.intellij.psi.impl.source.PsiClassReferenceType;
 import com.redhat.devtools.intellij.lsp4mp4ij.psi.core.java.diagnostics.JavaDiagnosticsContext;
 import com.redhat.devtools.intellij.lsp4mp4ij.psi.core.java.validators.JavaASTValidator;
 import com.redhat.devtools.intellij.lsp4mp4ij.psi.core.utils.PsiTypeUtils;
-import org.eclipse.lsp4j.DiagnosticSeverity;
 import com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.graphql.MicroProfileGraphQLConstants;
+import com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.graphql.TypeSystemDirectiveLocation;
+import org.eclipse.lsp4j.DiagnosticSeverity;
+import org.jetbrains.annotations.NotNull;
 
+import static com.redhat.devtools.intellij.lsp4mp4ij.psi.core.utils.AnnotationUtils.getAnnotation;
 import static com.redhat.devtools.intellij.lsp4mp4ij.psi.core.utils.AnnotationUtils.isMatchAnnotation;
 
 /**
  * Diagnostics for microprofile-graphql.
  *
+ * TODO: We currently don't check directives on input/output objects and their properties, because
+ * it's not trivial to determine whether a class is used as an input, or an output, or both. That
+ * will possibly require building the whole GraphQL schema on-the-fly, which might be too expensive.
+ *
  * @see https://download.eclipse.org/microprofile/microprofile-graphql-1.0/microprofile-graphql.html
  */
 public class MicroProfileGraphQLASTValidator extends JavaASTValidator {
 
-	private static final Logger LOGGER = Logger.getLogger(MicroProfileGraphQLASTValidator.class.getName());
+    private static final Logger LOGGER = Logger.getLogger(MicroProfileGraphQLASTValidator.class.getName());
 
-	@Override
-	public boolean isAdaptedForDiagnostics(JavaDiagnosticsContext context) {
-		Module javaProject = context.getJavaProject();
-		// Check if microprofile-graphql is on the path
-		return PsiTypeUtils.findType(javaProject, MicroProfileGraphQLConstants.QUERY_ANNOTATION) != null;
-	}
+    private static final String WRONG_DIRECTIVE_PLACEMENT = "Directive ''{0}'' is not allowed on element type ''{1}''";
 
-	@Override
-	public void visitMethod(PsiMethod node) {
-		validateMethod(node);
-		super.visitMethod(node);
-	}
+    @Override
+    public boolean isAdaptedForDiagnostics(JavaDiagnosticsContext context) {
+        Module javaProject = context.getJavaProject();
+        // Check if microprofile-graphql is on the path
+        return PsiTypeUtils.findType(javaProject, MicroProfileGraphQLConstants.QUERY_ANNOTATION) != null;
+    }
 
-	private void validateMethod(PsiMethod node) {
+    @Override
+    public void visitMethod(PsiMethod node) {
+        validateDirectivesOnMethod(node);
+        super.visitMethod(node);
+    }
 
-	}
+    @Override
+    public void visitClass(PsiClass node) {
+        validateDirectivesOnClass(node);
+        super.visitClass(node);
+    }
+
+
+    private void validateDirectivesOnMethod(PsiMethod node) {
+        for (PsiAnnotation annotation : node.getAnnotations()) {
+            // a query/mutation/subscription may only have directives allowed on FIELD_DEFINITION
+            if (isMatchAnnotation(annotation, MicroProfileGraphQLConstants.QUERY_ANNOTATION) ||
+                    isMatchAnnotation(annotation, MicroProfileGraphQLConstants.MUTATION_ANNOTATION) ||
+                    isMatchAnnotation(annotation, MicroProfileGraphQLConstants.SUBSCRIPTION_ANNOTATION)) {
+                validateDirectives(node, TypeSystemDirectiveLocation.FIELD_DEFINITION);
+            }
+        }
+        // any parameter may only have directives allowed on ARGUMENT_DEFINITION
+        for (PsiParameter parameter : node.getParameterList().getParameters()) {
+            validateDirectives(parameter, TypeSystemDirectiveLocation.ARGUMENT_DEFINITION);
+        }
+    }
+
+    private void validateDirectivesOnClass(PsiClass node) {
+        // a class with @GraphQLApi may only have directives allowed on SCHEMA
+        if(getAnnotation(node, MicroProfileGraphQLConstants.GRAPHQL_API_ANNOTATION) != null) {
+            validateDirectives(node, TypeSystemDirectiveLocation.SCHEMA);
+        }
+        // if an interface has a `@Union` annotation, it may only have directives allowed on UNION
+        // otherwise it may only have directives allowed on INTERFACE
+        if (node.isInterface()) {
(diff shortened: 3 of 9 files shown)


src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/TypeSystemDirectiveLocation.java
```diff
@@ -0,0 +1,15 @@
+package com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.graphql;
```
angelozerr, line 1: Add Header

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/TypeSystemDirectiveLocation.java
```diff
@@ -0,0 +1,15 @@
...
+
+public enum TypeSystemDirectiveLocation {
```
angelozerr, line 3: Add Javadoc and if it is possible add a link to the doc which defines those constants.

angelozerr:
Thanks so much @jmartisk for your great contribution.

I'm a newbie with GraphQL and as you changed the validation rule, I wonder if we should keep the ols validation rule by checking the GraphQL version. @jmartisk, @fbricon  what do you think about that?

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/TypeSystemDirectiveLocation.java
jmartisk: done

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/TypeSystemDirectiveLocation.java
jmartisk: done

jmartisk:
> I wonder if we should keep the ols validation rule by checking the GraphQL version. @jmartisk, @fbricon what do you think about that?

If it's common here to do different checks depending on the Quarkus version and you can point me to how I can reliably detect that version, then I'm fine with it

fbricon:
Quarkus version checks make sense, however we have no API in place for that yet. @jmartisk do you think you could add that?

jmartisk:
I'll look into it (but any pointers about how to do it are welcome)

jmartisk:
I've re-added the void check, plus some logic for determining the Quarkus version, but I'm really not sure if this is a good solution (probably isn't), so please have a look

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
```diff
@@ -14,65 +14,212 @@
...
+        // otherwise it may only have directives allowed on INTERFACE
+        if (node.isInterface()) {
+            if(getAnnotation(node, MicroProfileGraphQLConstants.UNION_ANNOTATION) != null) {
+                validateDirectives(node, TypeSystemDirectiveLocation.UNION);
+            } else {
+                validateDirectives(node, TypeSystemDirectiveLocation.INTERFACE);
+            }
+        }
+        // an enum may only have directives allowed on ENUM
+        if (node.isEnum()) {
```
angelozerr, line 133: `else if`

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
```diff
@@ -14,65 +14,212 @@
...
+    boolean allowsVoidReturnFromOperations = true;
+
+
+    @Override
+    public boolean isAdaptedForDiagnostics(JavaDiagnosticsContext context) {
+        Module javaProject = context.getJavaProject();
+        if(PsiTypeUtils.findType(javaProject, MicroProfileGraphQLConstants.QUERY_ANNOTATION) == null) {
+            return false;
+        }
+        String quarkusVersion = QuarkusModuleUtil.getQuarkusVersion(context.getJavaProject());
```
angelozerr, line 71: For the moment we don't need the getQuarkusVersion, we need the method `isQuarkusVersion(Module javaProject, String version)`

So I suggest:

 * set getQuarkusVersion as private
 * introduce a new new method `isQuarkusVersion(Module javaProject, String version)`

So you could just write after that:

`boolean allowsVoidReturnFromOperations = QuarkusModuleUtil.isQuarkusVersion(context.getJavaProject(), "3.1"));`

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/TypeSystemDirectiveLocation.java
angelozerr: Thanks!

projects/lsp4mp/projects/maven/microprofile-graphql/pom.xml
```diff
@@ -39,6 +39,11 @@
         <artifactId>smallrye-graphql-client-api</artifactId>
         <version>1.0.2</version>
     </dependency>
+    <dependency>
```
angelozerr, line 42: Perhaps it should be better to have 2 projects microprofile-graphqwith Quarkus 3.1 and Quarkus <3.1 since we have different validation kind?

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
jmartisk: I'm not following here, we need to check a range (`<3.1`), not one particular version number, so with `isQuarkusVersion` we would have to list all versions after 3.1, many of which aren't known yet. 
Even if we inverted the condition and then enumerated all versions BEFORE 3.1, we can't be 100 % sure that there won't ever be a 2.17 release, for example.

src/main/java/com/redhat/devtools/intellij/quarkus/QuarkusModuleUtil.java
```diff
@@ -179,6 +182,28 @@ public static boolean isQuarkusDeploymentLibrary(@NotNull LibraryOrderEntry libr
...
+    public static String getQuarkusVersion(Module module) {
+        Optional<VirtualFile> quarkusCoreJar = Arrays.stream(ModuleRootManager.getInstance(module).orderEntries()
+                        .runtimeOnly()
+                        .classes()
+                        .getRoots())
+                .filter(f -> Pattern.matches(QUARKUS_CORE_PATTERN.pattern(), f.getName()))
+                .findFirst();
+        if (quarkusCoreJar.isPresent()) {
+            Matcher matcher = QUARKUS_CORE_PATTERN.matcher(quarkusCoreJar.get().getName());
+            System.out.println("matches: " + matcher.matches());
```
angelozerr, line 200: Remove System.out or use LOGGER

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
angelozerr: > I'm not following here, we need to check a range (<3.1)

You are right, sorry I read too quickly your code. So My suggestion was to have an utility method like 

`QuarkusModuleUtil.isQuarkusVersionGreaterThan(String version)`  I'm pretty sure that we will need this utility method for another features.

angelozerr:
@jmartisk for your information we have the intention to create a release this week (perhaps thursday). See millestione at https://github.com/redhat-developer/intellij-quarkus/milestone/41

We includeded your PR for now, but if you need some more time, we will register to an another millestone.

I think your great PR is almost ready. Please fix my comments.

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
jmartisk: And what should that return if we can't detect the version at all (there might not be a Quarkus dependency at all) or it's a custom version, like 999-SNAPSHOT, where we can't reason about anything? I assume both of these cases should return false, right?

jmartisk: One more issue is how do you define "greater than"?  It might not be a sufficient determinator in all cases
For example:
- 4.0 does not have a feature
- 4.1 has it
- 3.0 doesn't have it
- 3.1 has it

Then, how would you check whether a particular version has this feature, using `isQuarkusVersionGreaterThan`? You can't use `isQuarkusVersionGreaterThan(3.1)` because then it would return `true` for 4.0, which is wrong. 
This approach, while it is convenient, has a lot of limitations IMO...

src/main/java/com/redhat/devtools/intellij/quarkus/QuarkusModuleUtil.java
jmartisk: Fixed

jmartisk:
Right, I'm just still a bit unsure about the approach you're suggesting regarding Quarkus version comparisons. See my comments above. In any case, I assume this is something we can always change later and it should not block getting this PR in.

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
angelozerr: I understand what you mean, but it would be very nice to have an utility method which check your usecases with version. I would like to avoid copy/paste code 

```
if(quarkusVersion != null) {
            Matcher matcher = QuarkusModuleUtil.QUARKUS_STANDARD_VERSIONING.matcher(quarkusVersion);
            if(matcher.matches()) { // ignore if this is not a standard version (for example, a custom snapshot)
```

for an another check version for another feature.

So I sggest that you define a method like

`boolean checkQuarkusVersion(Project project, Predicate<Matcher> versionMatcher)`

After that you could use it like this:

```
checkQuarkusVersion(context.getJavaProject(), matcher -> Integer.valueOf(matcher.group(1)) < 3 ||
                        (Integer.valueOf(matcher.group(1)) == 3) && Integer.valueOf(matcher.group(2)) == 0));
```

jmartisk: Ok that makes sense, I'll do that in a bit

jmartisk: Done, please check

angelozerr:
@jmartisk thanks so much for your  contribution (and your patience)! Your code is very good.

If CI build will success, we will able to merge your PR.

angelozerr:
@jmartisk the `testIncorrectDirectivePlacement ` test fails in CI build :

```
com.redhat.devtools.intellij.lsp4mp4ij.psi.core.graphql.java.MicroProfileGraphQLValidationTest

  Test testIncorrectDirectivePlacement FAILED (7.8s)

  java.lang.AssertionError: Unexpected diagnostics:
  [Diagnostic [
    range = Range [
      start = Position [
        line = 38
        character = 4
      ]
      end = Position [
        line = 38
        character = 15
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'FIELD_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 48
        character = 50
      ]
      end = Position [
        line = 48
        character = 61
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'ARGUMENT_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 72
        character = 59
      ]
      end = Position [
        line = 72
        character = 70
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'ARGUMENT_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 90
        character = 11
      ]
      end = Position [
        line = 90
        character = 15
      ]
    ]
    severity = Error
    code = Either [
      left = NO_VOID_QUERIES
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Methods annotated with microprofile-graphql's `@Query` cannot have 'void' as a return type."
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 94
        character = 11
      ]
      end = Position [
        line = 94
        character = 15
      ]
    ]
    severity = Error
    code = Either [
      left = NO_VOID_MUTATIONS
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Methods annotated with microprofile-graphql's `@Mutation` cannot have 'void' as a return type."
    tags = null
    relatedInformation = null
    data = null
  ]] expected:<[Diagnostic [
    range = Range [
      start = Position [
        line = 38
        character = 4
      ]
      end = Position [
        line = 38
        character = 15
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'FIELD_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 48
        character = 50
      ]
      end = Position [
        line = 48
        character = 61
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'ARGUMENT_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 72
        character = 59
      ]
      end = Position [
        line = 72
        character = 70
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'ARGUMENT_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ]]> but was:<[Diagnostic [
    range = Range [
      start = Position [
        line = 38
        character = 4
      ]
      end = Position [
        line = 38
        character = 15
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'FIELD_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 48
        character = 50
      ]
      end = Position [
        line = 48
        character = 61
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'ARGUMENT_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 72
        character = 59
      ]
      end = Position [
        line = 72
        character = 70
      ]
    ]
    severity = Error
    code = Either [
      left = WRONG_DIRECTIVE_PLACEMENT
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Directive 'io.openliberty.graphql.sample.Optimistic' is not allowed on element type 'ARGUMENT_DEFINITION'"
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 90
        character = 11
      ]
      end = Position [
        line = 90
        character = 15
      ]
    ]
    severity = Error
    code = Either [
      left = NO_VOID_QUERIES
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Methods annotated with microprofile-graphql's `@Query` cannot have 'void' as a return type."
    tags = null
    relatedInformation = null
    data = null
  ], Diagnostic [
    range = Range [
      start = Position [
        line = 94
        character = 11
      ]
      end = Position [
        line = 94
        character = 15
      ]
    ]
    severity = Error
    code = Either [
      left = NO_VOID_MUTATIONS
      right = null
    ]
    codeDescription = null
    source = "microprofile-graphql"
    message = "Methods annotated with microprofile-graphql's `@Mutation` cannot have 'void' as a return type."
    tags = null
    relatedInformation = null
    data = null
  ]]>
      at com.redhat.devtools.intellij.lsp4mp4ij.psi.core.graphql.java.MicroProfileGraphQLValidationTest.testIncorrectDirectivePlacement(MicroProfileGraphQLValidationTest.java:65)
```

jmartisk:
That should be hopefully fixed now

src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/graphql/java/MicroProfileGraphQLASTValidator.java
angelozerr: It is perfect, Thanks!

angelozerr:
> That should be hopefully fixed now

Thanks! Let's wait for CI build result.

angelozerr:
CI build is working now again. Thanks so much @jmartisk !

angelozerr:
@jmartisk for your information we release IJ Quarkus 1.28 yesterday which host your contribution. Thanks again.

PR #1170, opened 2023-09-15
fix: Qute fragments doesn't work with non inner Templates Java file

### LINKED ISSUE #1169: Qute fragments doesn't work with non inner Templates Java file
When i define a fragment in a `@CheckedTemplate` class the Code Lens shows `Create path/to/file.html` while the file exist. (attached image)

**Suggestion**

Depending on the situation it should do the following:

1. If the file does not exist, create the file, create the fragment in that file and jump to that file/fragment.
2. If the file does exist, but the fragment does not exist, create the fragment in that file and jump to that fragment.
3. If the file does exist and the fragment does exist just jump to that fragment.

image

angelozerr:
fix: Qute fragments doesn't work with non inner Templates Java file

Fixes #1169

Given this non inner class:

```java
package org.acme;

import io.quarkus.qute.CheckedTemplate;
import io.quarkus.qute.TemplateInstance;

@CheckedTemplate
public class Templates {

    public static native TemplateInstance auth$emailVerification(String email);
}

```
With this PR Fragment should be recognized 

image

### DIFF
--- a/projects/qute/projects/maven/qute-quickstart/src/main/java/org/acme/qute/ItemTemplates.java
+++ b/projects/qute/projects/maven/qute-quickstart/src/main/java/org/acme/qute/ItemTemplates.java
@@ -1,0 +1,15 @@
+package org.acme.qute;
+
+import io.quarkus.qute.CheckedTemplate;
+import io.quarkus.qute.TemplateInstance;
+
+import java.util.List;
+
+@CheckedTemplate
+public class ItemTemplates {
+
+    static native TemplateInstance items(List<Item> items);
+    static native TemplateInstance items$id1(List<Item> items);
+    static native TemplateInstance items3$id2(List<Item> items);
+    static native TemplateInstance items3$(List<Item> items);
+}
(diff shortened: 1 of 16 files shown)


angelozerr:
I need to write test.

angelozerr:
I have written tests, let's see if CI build is working.

fbricon:
looks like the open fragment codelens only open the file at Line 1, regardless of where the fragment's offset. 

I don't know how the links work, but maybe something like `editor.getCaretModel().moveToOffsetfragmentOffset)` could work?

angelozerr:
> looks like the open fragment codelens only open the file at Line 1, regardless of the fragment's offset.

Indeed it is because it is not implemented, see https://github.com/redhat-developer/intellij-quarkus/issues/1169#issuecomment-1720654298

> I don't know how the links work, but maybe something like editor.getCaretModel().moveToOffsetfragmentOffset) could work?

Perhaps but it requires that Qute Language server provides a command which returns the offset of a given fragment for a given template

angelozerr:
Tests are woking now.

fbricon:
Thanks @angelozerr!

PR #1173, opened 2023-09-18
fix: respect basePath in @CheckedTemplate to find templates

fbricon:
Still need to :
- handle corner cases, i.e empty paths, several "/"s
- write tests

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/qute/psi/internal/template/datamodel/CheckedTemplateSupport.java
+++ b/src/main/java/com/redhat/devtools/intellij/qute/psi/internal/template/datamodel/CheckedTemplateSupport.java
@@ -80,7 +80,8 @@ public class CheckedTemplateSupport extends AbstractAnnotationTypeReferenceDataM
         if (javaElement instanceof PsiClass) {
             PsiClass type = (PsiClass) javaElement;
             boolean ignoreFragments = isIgnoreFragments(checkedTemplateAnnotation);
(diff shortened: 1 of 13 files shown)


angelozerr:
@fbricon I suppose we have the same issue in JDT, right?

fbricon:
see https://github.com/redhat-developer/quarkus-ls/issues/893

src/main/java/com/redhat/devtools/intellij/qute/psi/internal/template/datamodel/CheckedTemplateSupport.java
```diff
@@ -112,15 +113,32 @@ private static boolean isIgnoreFragments(PsiAnnotation checkedTemplateAnnotation
         return ignoreFragment != null ? ignoreFragment.booleanValue() : false;
     }
 
+    public static String getBasePath(PsiAnnotation checkedTemplateAnnotation) {
```
angelozerr, line 116: Add Javdoc

angelozerr:
After adding Javadoc, I think we can merge this nice PR.

src/main/java/com/redhat/devtools/intellij/qute/psi/internal/template/datamodel/CheckedTemplateSupport.java
fbricon: done

angelozerr:
Thanks @fbricon !

PR #1177, opened 2023-09-22
fix: Qute support for multi module project doesn't work

### LINKED ISSUE #1164: Qute support for multi module project doesn't work
@ia3andy please be aware that multi module project with Qute doesn't work (you need to open the Qute project and not a project which could have one / several Qute module)

To fix this issue, it requires some work on Qute LS https://github.com/redhat-developer/quarkus-ls/issues/930

angelozerr:
fix: Qute support for multi module project doesn't work

Fixes #1164

Here a demo with https://github.com/gbourant/multi-module-template-test //cc @gbourant 

QuteMultiProjectIncludeInIJ

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,5 +5,5 @@ jetBrainsChannel=stable
 quarkusVersion=3.1.2.Final
 lsp4mpVersion=0.9.0
 quarkusLsVersion=0.16.0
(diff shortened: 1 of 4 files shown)


fbricon:
thanks @angelozerr!

PR #1185, opened 2023-09-28
feat: Check if project is a MicroProfile, Qute, etc project to map file with a language server

### LINKED ISSUE #850: Check if project is a MicroProfile, Qute, etc project to map file with a language server
Today a language server (Qute, Quarkus, etc) is started for any Java file, any properties file even if the project is not a Quarkus, Qute project.

We should improve the server extension point to add client custom filter to check if the project is a Quarkus, Qute project to know if the file must be connected to a language server.

It will fix seveal issues:

 * if project is not a Quarkus, Qute, etc project, the opened file will not connected to a language server (and it will avoid starting MicroProfile, Qute LS for any java files, any HTML files, any properties file)
 * for Telemetry, we will send an event only if project is a Qute / MicroProfile project, because today once you open a java file even i fit is not a Quarkus, Qute project, it send an event which says that project is a Qute, Quarku sproject.

It will fix for instance this kind of comment https://github.com/redhat-developer/intellij-quarkus/issues/735#issuecomment-1541545999

`also when the currently open project is not even a Quarkus application.`

### LINKED ISSUE #1121: Quarkus completion is available in all *.properties files in a Quarkus project
Quarkus completion is available in all *.properties files in a Quarkus project, but we just need it in application.properties, microprofile-config.properties

angelozerr:
feat: Check if project is a MicroProfile, Qute, etc project to map file with a language server

Fixes #850
Fixes #1121

This PR provides a new API `DocumentMatcher` which is used to match the document with a language server. This matcher is called if the file language is mapped with a language server. Once language is matched, teh document filter is processed to have more robust filter:

 * for properties file, we check the  name of properties files which fixes #1121
 * for Java file, we check that the file belong to a Quarkus / Qute project. It means that if you have a java project which is not a Qute / Quarkus project, language server should never started
 * for HTML files, we check that html files belongs to templates folder and belong to a Qute project.

The `DocumentMatcher` is used for the moment only when file is opened,it means if we add / remove a Qute / Quarkus dependency, it will change nothing. You will need to close all files and reoping it. @fbricon perhaps we could see that in an another PR?

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4ij/ContentTypeToLanguageServerDefinition.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4ij/ContentTypeToLanguageServerDefinition.java
@@ -1,17 +1,27 @@
 package com.redhat.devtools.intellij.lsp4ij;
 
 import com.intellij.lang.Language;
+import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.vfs.VirtualFile;
 
 import javax.annotation.Nonnull;
 import java.util.AbstractMap;
 
 public class ContentTypeToLanguageServerDefinition extends AbstractMap.SimpleEntry<Language, LanguageServersRegistry.LanguageServerDefinition> {
+
+    private final DocumentMatcher documentMatcher;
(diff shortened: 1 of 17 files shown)


angelozerr:
Close for https://github.com/redhat-developer/intellij-quarkus/pull/1188

PR #1220, opened 2023-10-09
build: add IC-2023.2.3 to the build matrix

angelozerr:
build: add IC-2023.2.3 to the build matrix

### DIFF
--- a/.github/workflows/IJ.yml
+++ b/.github/workflows/IJ.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        IJ: [IC-2022.2, IC-2022.3, IC-2023.1, IC-2023.2, IC-2023.2.1]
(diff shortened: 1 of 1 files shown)


fbricon:
2023.3 hasn't been released yet, you need to point to the EAP version

angelozerr:
> 2023.3 hasn't been released yet, you need to point to the EAP version

Indeed I created this PR when teh EAP will be released.

.github/workflows/IJ.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        IJ: [IC-2022.2, IC-2022.3, IC-2023.1, IC-2023.2, IC-2023.2.1]
```
fbricon, line 17: you should replace it with IC-2023.2.3 (released today) regardless of the 2023.3 EAP situation

angelozerr: thanks, fixed

PR #1226, opened 2023-10-16
fix: Non-public @Observes methods should be implicitly used

### LINKED ISSUE #1208: Non-public @Observes methods show as unused
Package protected @Observes methods should show as used, as they're valid, according to https://quarkus.io/guides/lifecycle#listening-for-startup-and-shutdown-events

Screenshot 2023-10-06 at 11 19 51

private methods work as well, although Quarkus shows a warning at startup:
> 2023-10-06 11:24:19,201 INFO  [io.qua.arc.pro.BeanProcessor] (build-14) Found unrecommended usage of private members (use package-private instead) in application beans:
	- Observer method com.redhat.devtools.stats.services.MarketPlaceStatisticsWatcher#onStart()
	
The check on public should be removed from https://github.com/redhat-developer/intellij-quarkus/blob/69541019c87167173e6781a316b8893a707fb757/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/core/JavaEEImplicitUsageProvider.java#L76

fbricon:
Fixes #1208


angelozerr:
Thanks @fbricon !

PR #1227, opened 2023-10-16
fix: show @Scheduled methods as implicitly used

### LINKED ISSUE #1201: @Scheduled methods should be marked as implicitly used
e.g.
```java
    @Scheduled(every = "2s")
    public void every2s() {
        System.err.println("Wake up!");
    }
```
Screenshot 2023-10-16 at 18 34 45

The class must be in the  `@ApplicationScoped` as per  https://quarkus.io/guides/scheduler#standard-scheduling

fbricon:
Fixes #1201


angelozerr:
Thanks @fbricon !

PR #1241, opened 2023-10-20
build: upgrade to Gradle 8.4 and build.gradle.kts

fbricon:


fbricon:
still need to adjust the workflows, then enable publishing from GHA

PR #1262, opened 2023-11-09
fix: Copy 'Run launch' tasks to the Maven run configuration

### LINKED ISSUE #1259: Run configuration 'Before launch' does not work
I don't know whether this is a bug in IntelliJ or the Quarkus plugin but the `Before launch` steps are not executed:

image

alerosmile:
The 'Before launch' tasks are configured in the QuarkusRunConfiguration. The must be added to the MavenRunConfiguration and GradleRunConfiguration, which are the ones which are actually executed.

Fixes #1259


alerosmile:
Fixes #1259

angelozerr:
Thanks @alerosmile !

LGTM, @fbricon @jeffmaury what do you think with this fix?

fbricon:
With this PR, the Build task is indeed executed before the launch.

I also tried the Open Browser task, to open http://localhost:8080, and this one runs **after** the launch, which is awesome! (maybe we could add a setting to add that task by default) 

Nice work @alerosmile, thanks!

PR #1268, opened 2023-11-20
fix: Connect Quarkus run configuration with Services view support

### LINKED ISSUE #1265: Connect Quarkus run configuration with Services view support
When Run button is clicked on Services view:

image

It starts the Quarkus run config, but:

 * it doesn't update the stop button
 * the right panel should display the console

image

angelozerr:
fix: Connect Quarkus run configuration with Services view support

Fixes #1265

Here a demo:

QuarkusAndServicesIJ2

Please note that link of App and Dev UI are in the same line than Running. I wanted to display thoses 2 links as Running child but it seems it is not possible.

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/run/dashboard/QuarkusRunDashboardCustomizer.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/run/dashboard/QuarkusRunDashboardCustomizer.java
@@ -1,0 +1,84 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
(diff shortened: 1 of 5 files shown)


src/main/java/com/redhat/devtools/intellij/quarkus/run/dashboard/QuarkusRunDashboardCustomizer.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public boolean updatePresentation(@NotNull PresentationData presentation, @NotNull RunDashboardRunConfigurationNode node) {
+        if (!(node.getConfigurationSettings().getConfiguration() instanceof QuarkusRunConfiguration)) {
+            return false;
+        }
+        node.putUserData(RunDashboardCustomizer.NODE_LINKS, null);
+        RunContentDescriptor descriptor = node.getDescriptor();
+        if (descriptor != null) {
+            ProcessHandler processHandler =  descriptor.getProcessHandler();
+            if (processHandler != null && !processHandler.isProcessTerminated()) {
```
fbricon, line 54: you also need to check if `io.quarkus:quarkus-vertx-http` is a dependency of the module, before showing links to the application url and dev ui.

It would be useless for a CLI app (try creating a quarkus project with only picocli as dependency)

angelozerr: fixed

PR #1270, opened 2023-11-23
fix: keep compatibility with 2023.3

fbricon:
MavenServerManager.WRAPPED_MAVEN was removed in 2023.3 without any deprecation warning, so we keep a local copy

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/buildtool/maven/MavenToolDelegate.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/buildtool/maven/MavenToolDelegate.java
@@ -93,8 +93,9 @@ public class MavenToolDelegate implements BuildToolDelegate {
             var distributionUrl = getWrapperDistributionUrl(ProjectUtil.guessProjectDir(project));
             if (distributionUrl != null) {
                 String mavenHome = mavenSettings.getMavenHome();
-                if (!MavenServerManager.WRAPPED_MAVEN.equals(mavenHome)){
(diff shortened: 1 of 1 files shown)


src/main/java/com/redhat/devtools/intellij/quarkus/buildtool/maven/MavenToolDelegate.java
```diff
@@ -93,8 +93,9 @@ public void processImport(Module module) {
             var distributionUrl = getWrapperDistributionUrl(ProjectUtil.guessProjectDir(project));
             if (distributionUrl != null) {
                 String mavenHome = mavenSettings.getMavenHome();
-                if (!MavenServerManager.WRAPPED_MAVEN.equals(mavenHome)){
-                    mavenSettings.setMavenHome(MavenServerManager.WRAPPED_MAVEN);
+                String WRAPPED_MAVEN = "Use Maven wrapper"; //MavenServerManager.WRAPPED_MAVEN was removed in 2023.3 without any deprecation warning!
```
angelozerr, line 96: define a private static field

fbricon: that's what I did 1st, but intellij said it was better to keep it local

angelozerr: very strange? Ok as you wish.

angelozerr:
Thanks @fbricon !

PR #1284, opened 2023-12-08
chore: add gradle wrapper validation

sbouchet:
Add a gradle wrapper validation steps for security reasons.


fbricon:
@sbouchet thanks!

PR #1288, opened 2023-12-12
feat: Declare server icon with icon attribute

angelozerr:
feat: Declare server icon with icon attribute

This PR requires https://github.com/redhat-developer/lsp4ij/pull/40


src/main/java/com/redhat/devtools/intellij/quarkus/lang/QuarkusIcons.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.devtools.intellij.quarkus.lang;
```
fbricon, line 1: header

fbricon:
thx @angelozerr

PR #1291, opened 2023-12-13
feat: use server bundle

angelozerr:
feat: use server bundle


fbricon (approved):
OK to merge once https://github.com/redhat-developer/lsp4ij/pull/43 is merged

PR #1298, opened 2023-12-19
build: replace remark with flexmark

fbricon:
Replaces obsolete, unmaintained, unsecure remark with flexmark.

Fixes an issue with documentation always being returned as plaintext instead of markdown.

Works best with https://github.com/redhat-developer/lsp4ij/pull/59


angelozerr:
Thanks @fbricon !

PR #1328, opened 2024-04-30
feat: add support for @Startup healthcheck diagnostic

fbricon:
See https://github.com/eclipse/lsp4mp/issues/443


angelozerr:
It works great, thanks so much @fbricon !

PR #1336, opened 2024-05-21
build: increase IJ dependency to 2023.2

fbricon:


angelozerr:
Thanks so much @fbricon !

PR #1342, opened 2024-06-13
fix: adapt the new API of LSP4IJ

angelozerr:
fix: adapt the new API of LSP4IJ

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/lsp/QuarkusServer.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/lsp/QuarkusServer.java
@@ -14,6 +14,7 @@ import com.intellij.ide.plugins.IdeaPluginDescriptor;
 import com.intellij.ide.plugins.PluginManagerCore;
 import com.intellij.openapi.extensions.PluginId;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.vfs.VirtualFile;
 import com.redhat.devtools.lsp4ij.server.JavaProcessCommandBuilder;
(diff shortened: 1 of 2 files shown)


angelozerr:
@aparnamichael @turkeylurkey when you will use LSP4IJ 0.0.2 (which will release soon) please do this change too in your side.

mrglavas:
@angelozerr I see this corresponds to a change to an interface (StreamConnectionProvider) in LSP4iJ. This doesn't look like it's binary compatible. It seems that if we had released our plug-in based on 0.0.1 it might have been broken by this change. Is LSP4iJ's API going to be stable after 0.0.2?

angelozerr:
> LSP4iJ's API going to be stable after 0.0.2?

I hope so, but as we have more and more people who use LSP4IJ, we have more and more use cases and refactoring is required.

I'm trying to avoid breaking API but as the project is on the start step, I think it is better to break API.

fbricon:
@mrglavas we were also aware you guys didn't have released your plugin yet, and no one else was using it, so we deemed it an acceptable change. Had you released against 0.0.1, we'd have stuck with the old code for a while longer.

yeekangc:
The team has been actively working to adopt a release of LSP4IJ formally.  We are looking to do so in an upcoming release in the coming quarter as soon as possible.  While we will manage the churn here, as per the discussion in this issue and else, it will be good to work out an approach that is more predictable and less disruptive when it comes to breaking changes.  Let us continue the discussion and work out what's best for LSP4IJ and the community.  Thank you.

PR #1344, opened 2024-06-24
fix: relax @Channel conditions for completion/validation

fbricon:
First pass at fixing the [Unrecognized property mp.messaging.* when Channel annotation is used along with Multi](https://github.com/redhat-developer/vscode-quarkus/issues/750) issue opened in VS Code Quarkus


fbricon:
It seems the whole messaging-related annotation support could be overhauled, as we apparently also need to allow `@Incoming` annotations on methods, e.g. in [one-two-three](https://github.com/ia3andy/one-two-three-quarkus/blob/main/dashboard/src/main/resources/application.properties#L7):

Screenshot 2024-07-08 at 11 11 40

PR #1347, opened 2024-06-29
fix: Formatting is broken when using simple {}

### LINKED ISSUE #1345: Formatting is broken when using  simple `{}`
as an example try https://github.com/quarkusio/quarkus-web-lab/tree/main/solution/1-cms

Open src/main/resources/templates/Cms/index.html and reformat.
Remove the `{}` and it works:

https://github.com/redhat-developer/intellij-quarkus/assets/2223984/dd2312e0-b591-4719-b586-3bcb6b7946ec

angelozerr:
fix: Formatting is broken when using simple {}

Fixes #1345


angelozerr:
@ia3andy we have a bug with the Qute IntelliJ parser that I have fixed (need to play more with this fix and need to write tests for this Qute IntelliJ parser).

The fix allows to consider that Qute section tag are closed correctly and now it is considered as HTML Text node. 

With the fix:, HTML element should be correctly indented, but as Qute section as considered as HTML Text node, the section cannot be indented, is it annoying?

Given this Qute template:

```html
<div>
<div>
                    {#if}
    {#if}
                {inject:flash.get("key")}
<form>
        </form>
        {/if}
                    {/if}
                        </div>
    </div>
```

After formatting:

```html
<div>
    <div>
        {#if}
        {#if}
        {inject:flash.get("key")}
        <form>
        </form>
        {/if}
        {/if}
    </div>
</div>
```

 * HTML elements are correctly indented:
 * Qute section are indented by using the HTML text node strategy format. As you can see all Qute section are idented with the same indent. Is it enough @ia3andy ?

Indenting Qute section is not trivial to do, we could do that in an another PR, but @ia3andy do you think it is enough for you?

ia3andy:
It's perfect 🤩 please make it available for 8 of July :)

angelozerr:
@ia3andy you can play with this PR by downloading thezip at  https://github.com/redhat-developer/intellij-quarkus/actions/runs/9722616563, see expalanation at https://github.com/redhat-developer/intellij-quarkus?tab=readme-ov-file#testing-the-ci-builds

It should install https://plugins.jetbrains.com/plugin/23257-lsp4ij/edit too I think otherwise you will need to install it.

ia3andy:
ok will there be a release before 8 of july? (I have a lab)

ia3andy (approved):
Thanks!!!!!

It's amazing :)

https://github.com/quarkusio/quarkus-web-lab/commit/0e78d8f5c9c2ac5c2c0b346845738f3139d277df

angelozerr:
> It's amazing :)

That's great! Thanks @ia3andy for your patience and to have tested the IJ plugin of my PR.

PR #1351, opened 2024-07-04
fix: allow inject: with constructor class which have arguments

angelozerr:
Open https://github.com/ia3andy/one-two-three-quarkus/blob/main/dashboard/src/main/resources/templates/main.html

You will see a false positive error with `inject:csrf`:

image

`inject:csrf` comes from https://github.com/quarkusio/quarkus/blob/e46f2ec5588858525634e42fa1c6146162ade461/extensions/resteasy-reactive/rest-csrf/runtime/src/main/java/io/quarkus/csrf/reactive/runtime/CsrfTokenParameterProvider.java#L13

This class is flagged because it has constructor arguments https://github.com/quarkusio/quarkus/blob/e46f2ec5588858525634e42fa1c6146162ade461/extensions/resteasy-reactive/rest-csrf/runtime/src/main/java/io/quarkus/csrf/reactive/runtime/CsrfTokenParameterProvider.java#L27

Quarkus manages this arguments injection, so the  `inject:csrf` must be valid.

This PR fix the issue:
 
 *   `inject:csrf` doesn't report error
 *   completion for  `inject:csrf` should work now:

image

//cc @ia3andy

### DIFF
--- a/projects/qute/projects/maven/quarkus3/src/main/java/org/acme/Bean3.java
+++ b/projects/qute/projects/maven/quarkus3/src/main/java/org/acme/Bean3.java
@@ -1,0 +1,9 @@
+package org.acme;
+
+public class Bean3 {
+
+    // Quarkus can inject Bean1 in the constructor without declaring the bean1 parameter with @Inject
+    public Bean3(Bean1 bean1) {
+
+    }
+}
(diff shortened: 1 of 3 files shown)


angelozerr:
@fbricon this PR has now a test.

angelozerr:
The CI build fails because of integration tests, but I think we can merge it.

fbricon:
Thanks @angelozerr! Now the same fix needs to be applied in the JDT implem.

angelozerr:
> Now the same fix needs to be applied in the JDT implem.

See https://github.com/redhat-developer/quarkus-ls/pull/964

PR #1365, opened 2024-08-08
feat: New Feature: outline extension BuildItems

### LINKED ISSUE #1094: New Feature: outline extension `BuildItem`s
When writing a Quarkus extension, it would be nice if the developer had a list of the existing BuildItems they can use. 

For example, if I want to deploy a `ServletContainerInitializer`, I'd use a `ServletContainerInitializerBuildItem`, so a search mechanism would be nice to have  

We have a page generated at https://quarkus.io/guides/all-builditems however it's a bit annoying to find what you need there. It would be nice if we had a list of potential builditems in an IDE tool window (bonus points if including those that are not available in the project)

angelozerr:
feat: New Feature: outline extension BuildItems

Fixes #1094

Here a demo with the current PR:

BuildItemWorkspaceSymbols

Here how it works:

 * you need to open a Java file to start the MicroProfile Language Server (it doesn"t work if the language server is not started). To fix this issue, LSP4IJ must be improved with https://github.com/redhat-developer/lsp4ij/issues/439
 * you need to select `Project and libraries` in the scope to retrieve BuildItem from the JARs
 * As symbols is used for a lot of things like Java class, I decided to fill the list with BuildItem which starts with `&`

@gastaldi what do you think about that and the demo? Is it enough? Do you need more things?

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/psi/internal/builditems/QuarkusBuildItemUtils.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/psi/internal/builditems/QuarkusBuildItemUtils.java
@@ -1,0 +1,53 @@
+package com.redhat.devtools.intellij.quarkus.psi.internal.builditems;
+
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.progress.ProgressIndicator;
+import com.intellij.psi.PsiClass;
+import com.intellij.psi.PsiModifier;
+import com.intellij.psi.search.ProjectScope;
+import com.intellij.psi.search.searches.DefinitionsScopedSearch;
+import com.redhat.devtools.intellij.lsp4mp4ij.psi.core.utils.PsiTypeUtils;
+import com.redhat.devtools.intellij.quarkus.QuarkusConstants;
+import com.redhat.microprofile.psi.internal.quarkus.renarde.java.RenardeConstants;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+public class QuarkusBuildItemUtils {
+
+    private QuarkusBuildItemUtils() {
+
+    }
+
+    /**
+     * Returns a set of all classes in the given project that extend Renarde's
+     * <code>Controller</code> class.
+     *
+     * @param project the project to search in
+     * @param monitor the progress monitor
+     * @return a set of all classes in the given project that extend Renarde's
+     * <code>Controller</code> class
+     */
+    public static Set<PsiClass> getAllBuildItemClasses(Module project, ProgressIndicator monitor) {
+        PsiClass buildItemType = PsiTypeUtils.findType(project, QuarkusConstants.QUARKUS_BUILD_ITEM_CLASS_NAME);
+        if (buildItemType == null) {
+            return Collections.emptySet();
+        }
+
+        Set<PsiClass> types = new HashSet<>();
+        DefinitionsScopedSearch.search(buildItemType, ProjectScope.getAllScope(project.getProject()), true)
+                .forEach(element -> {
+                    if (element instanceof PsiClass type && isValidBuildItem(type)) {
+                        types.add(type);
+                    }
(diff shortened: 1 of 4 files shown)


gastaldi:
@angelozerr that looks really cool, good job! Can we somehow also show the javadoc next to the class name (that explains what the build item is for)?

angelozerr:
> @angelozerr that looks really cool, good job! 

Thanks!

> Can we somehow also show the javadoc next to the class name (that explains what the build item is for)?

I fear that it is not possible by using standard `Symbols` tab. To do that I think we should develop specific UI, but I don't know if it is easy to do that.

fbricon:
My fear is about the need to use a specific prefix for triggering the builditem search, in this cas `&`. It's seems arbitrary/unnatural and would need to be documented extensively. I don't see a way to make it easily discoverable.

src/main/java/com/redhat/devtools/intellij/quarkus/psi/internal/builditems/QuarkusBuildItemUtils.java
```diff
@@ -0,0 +1,53 @@
+package com.redhat.devtools.intellij.quarkus.psi.internal.builditems;
```
fbricon, line 1: header

src/main/java/com/redhat/devtools/intellij/quarkus/psi/internal/builditems/QuarkusBuildItemUtils.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Set;
+
+public class QuarkusBuildItemUtils {
+
+    private QuarkusBuildItemUtils() {
+
+    }
+
+    /**
+     * Returns a set of all classes in the given project that extend Renarde's
```
fbricon, line 24: wrong doc

angelozerr:
> My fear is about the need to use a specific prefix for triggering the builditem search, in this cas &. It's seems arbitrary/unnatural and would need to be documented extensively. I don't see a way to make it easily discoverable.

The main problem is that I contribute to Symbol and it is difficult to retrieve BuildItem without special characters because Symbol return Java class. For JAX-REST we use special character`@`

angelozerr:
@gastaldi do you think we can merge this feature:

* I decided to use `&` to search BuildItem, do you like that? Doyou prefer an another character?
* because of limitation of LSP WorkspaceSymbol and IJ too, we cannot shows Javadoc, so do you think it is helpfull to have this feature?

gastaldi:
@angelozerr I agree with @fbricon's concerns here that using `&` seems arbitrary and unnatural. Although I understand the reason why it works that way, I think that having a separate window (listing the existing build items along with their description) would be more user-friendly.

angelozerr:
> , I think that having a separate window (listing the existing build items along with their description) would be more user-friendly.

Ok it wil require more work because it will require to develop a custom UI to show Javadoc.

ia3andy:
@gastaldi could you elaborate on the need for this? using  class search with `BuildItem` gives the same result and its easier to find

ia3andy:
@gastaldi at some point, I was thinking about using AI to generate a better all BuildItems doc:
- with a generated doc when it's not documented
- with examples of usage from the Quarkus codebase
- showing what's internal and what's not

gastaldi:
As I reported in https://github.com/redhat-developer/intellij-quarkus/issues/1094, the idea is to have a search mechanism to find the intended build item while developing an extension. Although searching for `BuildItem` may display the build items on the classpath, it doesn't help much, since you lack examples and you need the sources to understand what it does. 

We have a list of all build items with their respective javadocs in https://quarkus.io/guides/all-builditems, that's what I use when trying to find a match for a build item when developing an extension, but having it easily findable in the IDE would be easier

ia3andy:
@gastaldi what about introducing another plugin for Quarkus extension developers then (and not making the one for users heavier)?

PR #1369, opened 2024-08-20
chore(skip-release): Prepare for 2.0.0

angelozerr:
chore(skip-release): Prepare for 2.0.0

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -6,6 +6,17 @@
     </vendor>
 
     <change-notes><![CDATA[
+    <h3>2.0.0</h3>
+    <ul>
+      <li>Delegate "LSP support" for Microprofile / Quarkus / Qute to <a href="https://github.com/redhat-developer/lsp4ij">LSP4IJ</a></li>
+      <li>Workspace symbols support for JAX-RS endpoints </li>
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -6,6 +6,17 @@
     </vendor>
 
     <change-notes><![CDATA[
+    <h3>2.0.0</h3>
+    <ul>
+      <li>Delegate "LSP support" for Microprofile / Quarkus / Qute with <a href="https://github.com/redhat-developer/lsp4ij">LSP4IJ</a></li>
```
fbricon, line 11: s/with/to

src/main/resources/META-INF/plugin.xml
```diff
@@ -6,6 +6,17 @@
...
+      <li>Workspace symbols support for JAX-RS endpoints </li>
+      <li>Resolve system properties/environment variables while browsing the application.properties values</li>
```
fbricon, line 13: ... while browsing application.properties ...

src/main/resources/META-INF/plugin.xml
angelozerr: fixed.

src/main/resources/META-INF/plugin.xml
angelozerr: fixed

PR #1370, opened 2024-08-21
fix: set platformOnly to false to collect all extensions in the wizard

angelozerr:
Unselect by default the platform checkbox to retrieve for instance langChain4j qurkus extensions:

image


fbricon:
Thanks @angelozerr!

PR #1378, opened 2024-08-29
fix: Qute - incorrect behavior for nested template records

angelozerr:
fix: Qute - incorrect behavior for nested template records


src/main/java/com/redhat/devtools/intellij/qute/psi/internal/template/datamodel/CheckedTemplateSupport.java
```diff
@@ -254,6 +269,13 @@ private static void collectDataModelTemplateForCheckedTemplate(PsiClass type,
         }
     }
 
+    public static @Nullable String getParentClassName(PsiClass type) {
+        boolean innerClass = type.getContainingClass() != null;
+        String className = !innerClass ? null
+                : PsiTypeUtils.getSimpleClassName(type.getContainingFile().getName());
+        return className;
```
fbricon, line 276:
```java
if (type.getContainingClass() != null) {
   return PsiTypeUtils.getSimpleClassName(type.getContainingFile().getName());
}
return null;
```
is simpler to read IMO

fbricon (changes requested):

src/main/java/com/redhat/devtools/intellij/qute/psi/internal/template/datamodel/CheckedTemplateSupport.java
angelozerr: fixed.

PR #1405, opened 2024-10-18
fix: use a task execution facade for Gradle task execution

honourosis:
The current implementation of task execution in `AbstractGradleToolDelegate` strictly relies on the internals of the External System API. This is not necessary, as there is a correspondingly simple way to execute tasks. 

In addition, some of the used methods will be annotated as `@Internal` and will not be available for use by third-party plugins.


angelozerr:
Thanks @honourosis 

I don't know this code base, so let's me some time to test your PR but it looks good.

honourosis:
@angelozerr Are there any updates to this pull request? It looks like macos-runner was broken when the pipeline was running.

angelozerr:
LGTM.

Thanks @honourosis !

PR #1429, opened 2025-01-23
feat: Add error submitter for user feedback

### LINKED ISSUE #591: Implement ErrorReportSubmitter for better user feedback

pablomxnl:
Closes #591 

It uses an [issue template form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository), on the yaml, `projects` list is optional.

error_reporting_dialog

When the user clicks the Create GitHub Issue button the browser is launched and takes away the focus of the IDE.

github_issue

### DIFF
--- a/.github/ISSUE_TEMPLATE/error_report.yml
+++ b/.github/ISSUE_TEMPLATE/error_report.yml
@@ -1,0 +1,47 @@
+name: Quarkus Tools Plugin Error Report
+description: "File a bug report for Quarkus Tool Plugin for IntelliJ"
+title: "bug: "
+labels:
+  - bug
+projects:
+  - redhat-developer/37
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for reporting this issue!! Please include as much information as possible in the Bug Description
+        **Please make sure to search for the title of this report** before submitting a new issue to avoid duplicates. 
+        Tip: You can attach images, videos, or log files by dragging it into the text box.
+  - id: report
+    type: textarea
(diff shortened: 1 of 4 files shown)


angelozerr:
Thanks so much!

@fbricon what do you think?

fbricon:
I'm worried this will potentially generate plenty of duplicate issues as it won't force users to check for existing issues

pablomxnl:
>  I'm worried this will potentially generate plenty of duplicate issues as it won't force users to check for existing issues

Understandable,  for my plugins I send these errors to sentry . Perhaps maybe it would be better to capture these errors through the telemetry plugin.

angelozerr:
I wonder if when the issue is created we could check that a similar issue (same title) exists?

pablomxnl:
> I wonder if when the issue is created we could check that a similar issue (same title) exists?

Probably using the gh api,  but then would have to deal with tokens in PasswordSafe, as far as I know there is no way to configure GitHub issue workflow to search while/before creating (GitLab does it this way) 

this is how the issue form would look like btw

issue_form

this is how GitLab display similar issues on the issue creation screen:
GitLab-Similar-Issues

Another option could be use a webhook on issue opened and search for similar issues and mark duplicates

pablomxnl:
interesting!!, will take a look. 
I see retrofit is available as transitive dependency from the telemetry plugin, 
would be ok to use that  or add the dependency explicitly ??

PR #1436, opened 2025-02-25
fix: fix unit test and integration test

sbouchet:
fixes #1435 
fixes #1343

PR #1440, opened 2025-03-20
fix: fix local dev when using local build of LSP4IJ

sbouchet:


sbouchet:
integration tests to be fixed in #1436

angelozerr:
Thanks so much @sbouchet !

PR #1446, opened 2025-04-18
feat: use Qute LS 0.22.0

angelozerr:
feat: use Qute LS 0.22.0


angelozerr:
Replaced with https://github.com/redhat-developer/intellij-quarkus/pull/1448

PR #1463, opened 2025-07-09
fix: Get project labels from a given project

angelozerr:
While fixing a bug in LSP4IJ with workspace symbols, I have noticed that workspace symbols is queried for all opened project (search of JAX-RS method in Symbol tab (Ctrl+N) is done for all opened project  and not for the current project.

This behavior is bad:

 * for feature: you see JAX-RS methods coming from other opened project which is wrong.
 * for performance, worlspace symbols is triggered for all opened project.

After debugging IntelliJ Quarkus, I have noticed the problem comes from ProjectLabelManager which uses all opened project instead of using the current project.

This PR fixes those 2 problems (features + performance).


angelozerr:
@dessina-devasia @turkeylurkey @TrevCraw @mrglavas I think you should have the same problem. Once I will merge this PR, I suggest that you apply the same fix for your Liberty Tools plugin.

PR #1488, opened 2025-10-22
build: update github actions to their latest version

fbricon:
update github actions to their latest version, hoping it will fix the frequent errors caused by `gradle/actions/wrapper-validation` in the nightly builds

### DIFF
--- a/.github/workflows/IJ-latest.yml
+++ b/.github/workflows/IJ-latest.yml
@@ -16,11 +16,11 @@ jobs:
     steps:
       # Check out current repository
       - name: Fetch Sources
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       # Set up Java environment for the next steps
       - name: Setup Java
-        uses: actions/setup-java@v4
+        uses: actions/setup-java@v5
         with:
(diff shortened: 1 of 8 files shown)


angelozerr:
Thanks @fbricon !

The thing that I don't understand is that LSP4IJ uses v4 an dit is working?

@sbouchet could you help us please.

sbouchet:
> Thanks @fbricon !
> 
> The thing that I don't understand is that LSP4IJ uses v4 an dit is working?
> 
> @sbouchet could you help us please.

That's maybe too much to just fix gradle validation errors, but that's ok. 
for lsp4ij, there is no gradle validation for the nightly, that's why it's not failling. Maybe that's a good idea to add it ... 

@fbricon : LGTM, however, to pass the validation test, you must change the build.gradle.kts file to something similar to this:
https://github.com/redhat-developer/lsp4ij/blob/682741e97b120dd92018854a2294ba269cf11d06/build.gradle.kts#L155

angelozerr:
It fails https://github.com/redhat-developer/intellij-quarkus/actions/runs/18841721831 -(

PR #1613, opened 2026-06-20
fix: Add Support of Gradle 9

### LINKED ISSUE #1512: Support Gradle 9
Currently the gradle task `listQuarkusDependencies` is not compitable with Gradle 9.

```
10:42:31 PM: Executing 'listQuarkusDependencies -c C:\Users\fouad\IdeaProjects\fouad-backend\362567369308420759.gradle.kts -q --console plain'…

Unknown command-line option '-c'.
```

See this note regarding upgrading to Gradle 9: https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#removal_of_custom_build_layout_options

> [Removal of custom build layout options](https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#removal_of_custom_build_layout_options)
> The ability to specify custom locations for key build files from the command line has been removed in Gradle 9.0.0. The following options, deprecated in Gradle 8.x, are no longer supported:
> 
> -c, --settings-file — Specify a custom location for the settings file
> 
> -b, --build-file — Specify a custom location for the build file
> 
> In addition, the buildFile property on the [GradleBuild](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.GradleBuild.html) task has been removed. This means it is no longer possible to set a custom build file path via the GradleBuild task.

- https://github.com/gradle/gradle/pull/33092

Eng-Fouad:
## Summary

Since Gradle 9, the extension fails to collect Quarkus deployment dependencies: it runs `listQuarkusDependencies -c <settingsFile>`, but the `-c` / `--settings-file` option (and the `rootProject.buildFileName` property it relied on) were **removed in Gradle 9**. As a result the task never launches and the Quarkus deployment JARs are not added to the module classpath. Works on Gradle 8, broken on Gradle 9.

Fixes redhat-developer/intellij-quarkus#1512

## Root cause

`AbstractGradleToolDelegate` generated a **custom build file** (a copy of the project build script + an appended task) and a **custom settings file** (with `rootProject.buildFileName = …`), then selected them with `-c <settingsFile>`. Gradle 9 removed:

- the `-c` / `--settings-file` (and `-b` / `--build-file`) command-line options, and
- `rootProject.buildFileName`.

## Fix

Collect the deployment dependencies through a Gradle **init script** (`--init-script`) instead, which is supported by both Gradle 8 and Gradle 9.

- The init script contributes the `quarkusDeployment` configuration, the deployment dependencies and the `listQuarkusDependencies` task to the project in an `afterEvaluate` hook.
- It no longer copies or modifies the user's `build.gradle(.kts)` / `settings.gradle(.kts)` — the project's own repositories are used to resolve the deployment artifacts.
- A single **Groovy** init script works for both Groovy and Kotlin DSL projects because it only uses the DSL-agnostic Gradle project API. As a result, `GradleGroovyToolDelegate` and `GradleKotlinToolDelegate` collapse to plain DSL (build-script name) detection.

## Tests

Adding `9.6.0` to the existing `GradleTestCase` parameter set is **not** a valid way to validate this on Gradle 9: every Gradle fixture applies the Quarkus Gradle plugin, and the plugin versions those fixtures pin (1.x/2.x) cannot configure on Gradle 9 themselves (e.g. the removed `JavaPluginConvention`, and an `integrationTestImplementation` configuration conflict). The project fails before `listQuarkusDependencies` ever runs, so the result would be a false signal — unrelated to this fix.

Instead:

- The integration-test suite stays on Gradle 8.x, with a comment in `GradleTestCase.data()` explaining why.
- A focused, **plugin-free** test, `GradleListQuarkusDependenciesInitScriptTest`, drives the real init script through the Gradle Tooling API on both **Gradle 8.6 and 9.6.0** and asserts that a deployment dependency's **binary and sources** JARs are resolved in the exact format the delegate parses.

| Check | Gradle 8.6 | Gradle 9.6.0 |
|---|---|---|
| `GradleListQuarkusDependenciesInitScriptTest` (binary + sources resolution) | ✅ | ✅ |
| `GradleQuarkusLibraryTest` (runs `listQuarkusDependencies` end-to-end) | ✅ | n/a (fixture plugin not Gradle 9 compatible) |

## Notes / follow-ups

- Running the full integration-test suite on Gradle 9 will require modernizing the Gradle fixtures to a Quarkus 3.x version that supports Gradle 9. That's a larger, separate change and is out of scope here.

## Changes

- `AbstractGradleToolDelegate`: generate and run a Gradle init script (`--init-script`); remove the custom build/settings-file generation and the now-unused abstract methods.
- `GradleGroovyToolDelegate` / `GradleKotlinToolDelegate`: reduce to build-script name + display metadata.
- `GradleTestCase`: keep the suite on Gradle 8.6, with an explanatory comment.
- New `GradleListQuarkusDependenciesInitScriptTest`: Gradle 9 (and 8) regression test for the init-script mechanism.

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/quarkus/buildtool/gradle/AbstractGradleToolDelegate.java
+++ b/src/main/java/com/redhat/devtools/intellij/quarkus/buildtool/gradle/AbstractGradleToolDelegate.java
@@ -80,9 +80,15 @@ public abstract class AbstractGradleToolDelegate implements BuildToolDelegate {
      * only relies on the DSL agnostic Gradle project API. This replaces the legacy approach based on a custom
      * settings file (selected with the {@code -c} option) and {@code rootProject.buildFileName}, both removed
      * in Gradle 9.
+     * <p>
+     * The configuration/dependencies/task are contributed to the project whose directory matches the analyzed
+     * module directory (not necessarily the root project), so the deployment artifacts are resolved with that
+     * module's own repositories. This keeps multi-module projects working, where repositories may be declared
+     * per subproject.
      * <ul>
      *     <li>{@code %1$s} is the list of deployment dependencies to resolve</li>
      *     <li>{@code %2$s} is the path of the file where the resolved artifacts are written</li>
+     *     <li>{@code %3$s} is the directory of the analyzed module (used to select the target Gradle project)</li>
      * </ul>
      */
     private static final String INIT_SCRIPT_TEMPLATE = """
@@ -91,7 +97,7 @@ public abstract class AbstractGradleToolDelegate implements BuildToolDelegate {
             import org.gradle.api.artifacts.result.ResolvedArtifactResult
 
             allprojects { proj ->
-                if (proj == proj.rootProject) {
+                if (proj.projectDir.canonicalPath == new File('%3$s').canonicalPath) {
                     proj.afterEvaluate { project ->
                         project.configurations.create('quarkusDeployment')
             %1$s\
@@ -172,7 +178,7 @@ public abstract class AbstractGradleToolDelegate implements BuildToolDelegate {
                                  @NotNull Set<String> deploymentIds,
                                  @NotNull List<VirtualFile>[] result) throws IOException {
         Path outputPath = Files.createTempFile(null, ".txt");
-        Path initScript = generateInitScript(outputPath, deploymentIds);
+        Path initScript = generateInitScript(outputPath, deploymentIds, getModuleDirPath(module));
         TaskCallback callback = new TaskCallback() {
             @Override
             public void onSuccess() {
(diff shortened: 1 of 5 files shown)


angelozerr:
Thanks so much @Eng-Fouad !

I am newbie with Gradle. Is your PR is ready?

Eng-Fouad:
> Thanks so much @Eng-Fouad !
> 
> I am newbie with Gradle. Is your PR is ready?

Yes it is ready.

angelozerr:
I trust you, but if there is some bugs or some improvment, are you OK to take care them?

Eng-Fouad:
> I trust you, but if there is some bugs or some improvment, are you OK to take care them?

Yes sure.

angelozerr:
Thanks @Eng-Fouad !

PR #1614, opened 2026-06-20
fix: Avoid duplicate Quarkus config property metadata

Eng-Fouad:
## Summary

Some Quarkus config properties are contributed to the metadata **twice**, which surfaces in the IDE as a "Go to definition" chooser with two identical targets (same file, same line) instead of navigating directly. It affects some config keys such as `quarkus.swagger-ui.oauth-client-id`, while other keys (e.g. `quarkus.oidc-client.discovery-enabled`) navigate fine.

aa

## Root cause

Some Quarkus config classes (e.g. `io.quarkus.swaggerui.deployment.SwaggerUiConfig`) are interfaces annotated with **both** `@ConfigRoot` and `@ConfigMapping`:

```java
@ConfigRoot
@ConfigMapping(prefix = "quarkus.swagger-ui")
public interface SwaggerUiConfig {
    Optional<String> oauthClientId();
    // ...
}
```

In `PropertiesManager`, the search query merges every provider's pattern, so such a class is matched by **both** the `@ConfigRoot` search (`QuarkusConfigRootProvider`) and the `@ConfigMapping` search (`QuarkusConfigMappingProvider`) and ends up being processed more than once. `PropertiesCollector.addItemMetadata` appended **every** contribution unconditionally, so the same property was added twice with identical `(name, sourceType, sourceField, sourceMethod)`.

This is exactly why only *some* properties duplicate: modern build-time configs use the `@ConfigRoot @ConfigMapping` interface style (matched by two providers), whereas many runtime configs carry only one of the two annotations (matched once).

## Fix

`PropertiesCollector` now skips a property whose `(name, sourceType, sourceField, sourceMethod)` was already collected. This collapses true duplicates while preserving:

- distinct properties from the same type (different method), and
- the same property name coming from **different** source members (no over-merging).

The change is confined to the per-property `addItemMetadata` path; the existing `merge(...)` strategies (`IGNORE_IF_EXISTS` / `REPLACE` / `FORCE`) are untouched.

## Tests

`PropertiesCollectorTest` gains three cases:

- `addItemMetadataIgnoresDuplicateForSameSourceMember` — identical contributions collapse to one.
- `addItemMetadataKeepsDistinctPropertiesFromSameType` — different methods on the same type are both kept.
- `addItemMetadataKeepsSameNameFromDifferentSourceMembers` — same name from different source types is kept.

Existing property tests still pass (e.g. `GradleQuarkusConfigRootHibernateRestEasyTest`, `MicroProfileConfigPropertyTest`), and they assert a property exists *exactly once*, so they'd fail if the dedup dropped a legitimate entry.

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/core/PropertiesCollector.java
+++ b/src/main/java/com/redhat/devtools/intellij/lsp4mp4ij/psi/internal/core/PropertiesCollector.java
@@ -15,9 +15,11 @@ package com.redhat.devtools.intellij.lsp4mp4ij.psi.internal.core;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
 import com.redhat.devtools.intellij.lsp4mp4ij.psi.core.IPropertiesCollector;
 import org.eclipse.lsp4mp.commons.MicroProfilePropertiesScope;
@@ -38,6 +40,12 @@ public class PropertiesCollector implements IPropertiesCollector {
 
 	private final Map<String, ItemHint> hintsCache;
 
+	/**
+	 * Keys of the properties already collected, used to avoid contributing the same property twice for
+	 * the same source member.
+	 */
+	private final Set<String> propertyKeys;
+
 	private final boolean onlySources;
 
 	public PropertiesCollector(ConfigurationMetadata configuration, List<MicroProfilePropertiesScope> scopes) {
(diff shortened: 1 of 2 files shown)


angelozerr:
@Eng-Fouad I don't understand when I try to see changes with github https://github.com/redhat-developer/intellij-quarkus/pull/1614/changes#diff-d098208cf5b52891bcddf71cc5040f29aa5b46c3d41723ebccf6f01c1906facf it says me that Binary file not shown:

image

Have you some idea?.

Eng-Fouad:
@angelozerr Fixed.

angelozerr:
LGTM, thanks @Eng-Fouad !

PR #1625, opened 2026-09-16
fix: Replace Query.findAll() for-loops with forEach lambda calls

anusreelakshmi934:


angelozerr:
Thanks @anusreelakshmi934 !

